PR #30, opened 2015-01-29
Spatial portal components

djtfmartin:
Spatial portal components

Includes
- layers service
- spatial portal UI
- layer-ingestion
- geoserver

fixes for vagrant 1.7 +
installation of postgres with postgis extensions (supported for ubuntu trusty + precise)
cleanup of vagrant inventories


ansible/inventories/vagrant/spatial-portal-vagrant
```diff
@@ -0,0 +1,13 @@
+[spatial-portal]
+vagrant1
+
+[spatial-portal:vars]
+is_vagrant=true
+local_repo_dir=~/.ala
+data_dir=/data
+tomcat_webapps=/var/lib/tomcat7/webapps
```
m-r-c: tomcat_webapps property is not required

ansible/roles/spatial-portal/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
+- include: ../../common/tasks/setfacts.yml
```
m-r-c: If you tag every task in this file with a 'spatial-portal' tag we can then run the playbook with that tag: this makes the playbook considerably faster if all the software dependencies (java, tomcat, db, etc etc) already exist by just running these tasks.

ansible/roles/layers-service/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
+- include: ../../common/tasks/setfacts.yml
+
+- name: install Postgis and GDAL Execs
```
m-r-c: Doesn't the postgresql role install postgis as well? Or this something different? 

It seems cleaner to have dedicated roles for software installs, so perhaps this could be moved to the postgresql role or a postgis role?

ansible/roles/layers-service/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+- name: copy transient files to tmp (schemas etc)
+  copy: src={{item}} dest=/tmp
+  with_items:
+    - layersdb.sql
+    - stored_procs.sql    
+  tags:
+    - createdb
+
+# Not working currently.....
+- name: setup the DB - WARNING!!! this isnt idempotent yet
```
m-r-c: There are dedicated postgresql tasks for creating databases and users. The problem with command tasks like this is that ansible always treats them as 'changed' so they run every time.

The postgresql_db and postgresql_user tasks are idempotent. They also allow you to register a variable to record whether they were executed or not: this can then be used to determine whether to execute the DB scripts below. E.g.:
```
- name: create database
  postgresql_db: name={{ volunteer_portal_db_name }}
  sudo_user: postgres
  sudo: True
  register: dbschema
  tags:
    - db
    - volunteer-portal
```
...

ansible/roles/layers-service/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+    - createdb
+
+- name: setup stored procedures - WARNING!!! this isnt idempotent yet
+  command: "psql -d {{ layers_db_name | default('layersdb') }} -a -f /tmp/stored_procs.sql"
+  sudo_user: postgres
+  sudo: True  
+  tags:
+    - createdb
+
+- name: setup the DB user password
```
m-r-c: ...you can then add a 'when' condition to this task so the script is only executed if the db was just created:
```
when: dbschema.changed
```

ansible/roles/layers-service/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+  command: "psql -d {{ layers_db_name | default('layersdb') }} -c \"ALTER USER {{ layers_db_username }} WITH PASSWORD '{{ layers_db_password }}' \""
+  sudo_user: postgres
+  sudo: True  
+  tags:
+    - createdb
+
+- name: download from maven repo
```
m-r-c: This task shouldn't be necesssary: tomcat_deploy below will download and deploy the WAR file.

ansible/roles/layers-service/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
+- include: ../../common/tasks/setfacts.yml
```
m-r-c: If you tag every task in this file with a 'layers-service' tag we can then run the playbook with that tag: this makes the playbook considerably faster if all the software dependencies (java, tomcat, db, etc etc) already exist by just running these tasks.

ansible/roles/layers-service/tasks/main.yml
djtfmartin: i did ponder this. The postgis extensions are now shipped in one pkg.
They are also harmless if not used.

ansible/roles/spatial-portal/tasks/main.yml
djtfmartin: good tip. I usually just run --skip-tags "java,tomcat" for the same effect.

ansible/roles/geoserver/tasks/main.yml
```diff
@@ -0,0 +1,11 @@
+- include: ../../common/tasks/setfacts.yml
```
m-r-c: If you tag every task in this file with a 'geoserver' tag we can then run the playbook with that tag: this makes the playbook considerably faster if all the software dependencies (java, tomcat, db, etc etc) already exist by just running these tasks.

ansible/layers-service.yml
```diff
@@ -0,0 +1,11 @@
+- hosts: layers-service
+  roles:
+    - java
+    - common
+    - tomcat
+    - apache
+    - geoserver
+    - postgresql    
+    - { role: postgresql_instance, db_name: "{{ layers_db_name | default('layersdb')}}",  db_username: "{{ layers_db_username | default('postgres') }}", db_password: "{{ layers_db_password | default('password') }}" }
```
m-r-c: Can the db configuration be moved into the layers-service role to be consistent with other playbooks?

ansible/AnsibleSkeleton.groovy
```diff
@@ -1,275 +0,0 @@
-class AnsibleSkeleton {
```
m-r-c: Can we leave this file in the repo for future use (generates the basic structure for new ansible scripts)?

We should remove ```tomcat_webapps={{tomcat_dir}}/webapps``` from line 242, and add ```groupId: ala.org.au``` at line 73, too.

ansible/roles/layers-service/tasks/main.yml
djtfmartin: yep. ta

ansible/roles/layers-service/templates/config/data.xml
```diff
@@ -0,0 +1,35 @@
...
+       http://www.springframework.org/schema/tx http://www.springframework.org/schema/tx/spring-tx-3.0.xsd
+       http://www.springframework.org/schema/aop http://www.springframework.org/schema/aop/spring-aop-3.0.xsd">
+
+    <tx:annotation-driven transaction-manager="transactionManager"
+                          proxy-target-class="true" />
+
+    <bean id="dataSource" class="com.mchange.v2.c3p0.ComboPooledDataSource"
+                                destroy-method="close"
+                                p:driverClass="org.postgresql.Driver"
+                                p:jdbcUrl="jdbc:postgresql://localhost:5432/{{ layers_db_name | default('layersdb')}}"
```
m-r-c: Should the db hostname be a property too, so we have the option of running the db on a different server?

ansible/roles/spatial-portal/templates/webportal-config.properties
```diff
@@ -0,0 +1,136 @@
+#CAS authentication related properties
+casProperties=casServerLoginUrl,serverName,centralServer,casServerName,uriFilterPattern,uriExclusionFilter,authenticateOnlyIfLoggedInFilterPattern,casServerLoginUrlPrefix,gateway,casServerUrlPrefix,contextPath
+disableCAS=false
+serverName={{spatial_portal_base_url | default('http://spatial.ala.org.au')}} 
+contextPath={{spatial_portal_context_path | default('http://spatial.ala.org.au')}}
```
m-r-c: Default value for contextPath shouldn't be the full url, should it?

djtfmartin: err no. Well spotted.

ansible/roles/spatial-portal/tasks/main.yml
djtfmartin: done

ansible/roles/geoserver/tasks/main.yml
djtfmartin: done

ansible/roles/layers-service/tasks/main.yml
djtfmartin: done

djtfmartin:
@m-r-c thanks for good feedback & suggestions. All incorporated.

m-r-c:
Looks good. Merged.

PR #36, opened 2015-02-11
Refine the ansible scripts for ala-collectory and generic collectory

linaizhong:


ansible/collectory-standalone.yml
```diff
@@ -1,8 +1,8 @@
...
-    - mysql
-    - tomcat    
-    - apache
-    - collectory
+    - { role: common, collectory: ala-collectory, collectory_version: 1.0 }
+    - { role: java, collectory: ala-collectory, collectory_version: 1.0 }
+    - { role: mysql, collectory: ala-collectory, collectory_version: 1.0 }
+    - { role: tomcat, collectory: ala-collectory, collectory_version: 1.0 }
+    - { role: apache, collectory: ala-collectory, collectory_version: 1.0 }
+    - { role: collectory, collectory: ala-collectory, collectory_version: 1.0 }
```
m-r-c: You only need to parameterise the collectory role. The others can remain as they were.

ansible/roles/collectory/templates/config/generic-collectory-config.properties
```diff
@@ -0,0 +1,73 @@
+# CAS Config
```
m-r-c: I don't think we need to have a specific file for the generic collectory. The base template should be sufficient for any collectory implementation: anything that needs to change between implementations (e.g. URLs) would be configured in the inventory file.

ansible/generic-collectory-standalone.yml
```diff
@@ -0,0 +1,8 @@
+- hosts: collectory
```
m-r-c: As above

ansible/collectory-standalone.yml
linaizhong: Every task (common, java, mysql, etc.) includes setfacts.yml that need to set the war url, which contains the application Id and Version. Our application Id and Version are variables. When being deployed, any task will stop working with the following error example if we did not pass the parameters:

TASK: [java | set remote repositories (ALA)] ********************************** 
fatal: [vagrant1] => One or more undefined variables: 'collectory' is undefined
FATAL: all hosts have already failed -- aborting

That is the reason I pass the parameters for all tasks.

m-r-c: Interesting. Looks like it's just the version that is used by common. This would explain why the biocache stuff uses "LATEST" for the version.

Rather than parameterising every role in the playbook, could you add a default value to the version variable in roles/collectory/vars/main.yml?
```
version: "{{ collectory_version | default('LATEST') }}"
```

linaizhong:
Only collectory scripts to deploy ala-collectory and generic-collectory applications.

linaizhong:
Please check if this is acceptable. Actually, there are two ways to put the parameters, one is to put the parameters in the collectory-standalone.yml, another is to put in the different configuration files under inventories/vagrant.

m-r-c:
Looks good. Merged.

PR #47, opened 2015-03-11
changes for APPD

djtfmartin:


ansible/roles/biocache-db/tasks/main.yml
```diff
@@ -2,17 +2,27 @@
...
+    - biocache_db
 
 - name: ensure cassandra is running
   wait_for: port=9160 delay=30
+  tags:
+    - biocache_db
 
 - name: create schema
   shell: "cassandra-cli < /tmp/cassandra-schema.txt"
+  tags:
```
m-r-c: What happens when this script is run against an existing DB? I think the ansible shell task executes every time, so you would blow away an existing DB. If that's the case, then you'll need to add a ```when``` condition to this task so it only runs if the schema does not already exist.

ansible/roles/biocache-hub/templates/config/config.properties
```diff
@@ -6,8 +6,8 @@ grails.resources.work.dir=/data/{{biocache_hub}}/cache
 casProperties=casServerLoginUrl,serverName,centralServer,casServerName,uriFilterPattern,uriExclusionFilterPattern,authenticateOnlyIfLoggedInFilterPattern,casServerLoginUrlPrefix,gateway,casServerUrlPrefix,contextPath
 serverName={{biocache_hub_base_url}}
 casServerName=https://auth.ala.org.au
-uriFilterPattern=/admin,/admin/.*
-uriExclusionFilterPattern=/occurrences/shapeUpload,/images.*,/css.*,/js.*,.*json,/help/.*
+uriFilterPattern={{ biocache_hub_uri_filter_pattern | default('/admin,/admin/.*') }}
```
m-r-c: Can you replace the hardcoded auth URLs with variables please?

ansible/roles/biocache-db/tasks/main.yml
djtfmartin: its harmless when ran against an existing DB.

PR #58, opened 2015-05-20
switchable name indexes, separate biocache properties

djtfmartin:
cc @nickdos and @m-hope 

this change has implications for APPD. This should allow switching the name index in use via the inventory.


m-r-c:
Looks good. Merged.

PR #66, opened 2015-06-24
Tomcat deploy task doesn't try to backup inexistent files anymore.

niconoe:
Hi guys,

When playing with the last version of the Ansible scripts, I noticed errors because Tomcat deploy launch the new backup task unconditionally. This fails the first time, since there's no war file to copy.

I seems this patch fix the issue, but I'm very new to Ansible.


djtfmartin:
Thanks a stack Nicolas !
I'll give it test....

m-r-c:
Oops. Didn't test that one very well, did I? Thanks for the fix - looks good.

PR #73, opened 2015-10-15
Changes as a result of ALS work

djtfmartin:
Includes roles for new BIE


ansible/roles/bie-hub/templates/bie-hub-config.properties
```diff
@@ -0,0 +1,33 @@
+bie.index.url={{ bie_index_url | default('http://localhost:8090/bie-index') }}
```
m-r-c: localhost?

djtfmartin: Im not sure what sensible default is. Ive changed to point to dev environment for now (in the absence of an ALA prod environment).

ansible/roles/biocache-properties/templates/biocache-config.properties
```diff
@@ -97,7 +97,7 @@ default.country={{ default_country | default('') }}
 sample.fields={{ sample_files | default('none') }}
 
 # The list of default fields to use if a list can not be obtained from the spatial layers.
-default.sample.fields=cl20,cl23
+default.sample.fields=
```
m-r-c: Do we still need the commented 'defaultFieldsToSample' block here?

ansible/roles/common/tasks/main.yml
```diff
@@ -8,26 +8,24 @@
 # On CSIRO VMs the mounted storage is typically on /srv/data and we need to create a symlink from /data -> /srv/data 
 # On Nectar & EC2 Ubuntu VMs this is usually /mnt  
 #
-- name: ensure local repository exists [happens locally]
-  action: file dest={{local_repo_dir}} state=directory force=yes
+# - name: ensure local repository exists [happens locally]
```
m-r-c: Remove if these are not needed any more.

ansible/roles/common/vars/main.yml
```diff
@@ -1,10 +1,10 @@
 # common variables across all roles
 
 # the application data directory
-data_dir_symlink: /data
+# data_dir_symlink: /data
```
m-r-c: Remove commented values

ansible/roles/jenkins/tasks/main.yml
```diff
@@ -62,14 +62,14 @@
...
 
-- name: enable the default ssl site
-  command: a2ensite default-ssl 
-  args:
-   creates: /etc/apache2/sites-enabled/default-ssl.conf
-  when: ansible_os_family == "Debian" 
-  notify: reload apache
-  tags: jenkins
-
+# Disabled - SSL certificates need to be configured properly
```
m-r-c: What is required for this?

ansible/roles/nameindex/tasks/main.yml
```diff
@@ -23,8 +23,15 @@
   tags:
     - "nameindex"  
 
+- name: set name index to use (Custom)
+  set_fact:
+    lucene_namematching_url: "{{ custom_namematching_url }}"
```
m-r-c: Is this variable being used somewhere?

ansible/roles/bie-hub/templates/bie-hub-config.properties
m-r-c: To be honest, I'm not a fan of having default values for properties like this. I would rather see a build error or a runtime failure if something isn't set than to have the system point to an unexpected service. That's a broader issue though.

ansible/roles/biocache-properties/templates/biocache-config.properties
djtfmartin: No. The default fields doesnt make sense for other installations

ansible/roles/jenkins/tasks/main.yml
djtfmartin: not sure i follow. This is being removed...

m-r-c: Sorry, I was referring to the comment "SSL certificates need to be configured properly". If we need to update the scripts to do this, then we should do it now. Otherwise, we should remove this block instead of commenting it out.

djtfmartin: I agree and agree in general with the "remove commented crap" prinicple. Will remove.

ansible/roles/nameindex/tasks/main.yml
djtfmartin: custom_namematching_url is a value that can be used to specify another lucene index to use. Im using the ALS and assume others will use for other deployments

PR #80, opened 2016-02-09
Change Ansible instructions to install from PIP

ansell:
The PPA that was referenced is currently hosting ansible-2.0, and may bump to ansible-2.1 in the near future. Checking out ansible-1.8.4 from Git is the safest way to ensure that it is used consistently instead of the newer versions.


ansell:
There may need to be something else happen, or I have a bug, as ansible doesn't run correctly after that sequence of operations, although it installed without any visible errors.

ansell:
After changing to installing from PIP, ansible-1.8.4 runs for me.

PR #95, opened 2016-03-15
Fixed default skin template assignment as generic. Added APPD support.

sadeghim:
Fixed default skin template assignment as generic.
Added APPD support.

### DIFF
--- a/ansible/roles/biocache-properties/tasks/main.yml
+++ b/ansible/roles/biocache-properties/tasks/main.yml
@@ -20,14 +20,12 @@
 
 - name: Download SDS layers
   get_url: url={{ sds_layers_url | default('http://biocache.ala.org.au/archives/layers/sds-layers.tgz') }} dest="{{data_dir}}/biocache/layers/sds-layers.tgz" force=yes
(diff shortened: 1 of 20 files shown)


ansible/roles/biocache-properties/tasks/main.yml
```diff
@@ -20,14 +20,12 @@
 
 - name: Download SDS layers
   get_url: url={{ sds_layers_url | default('http://biocache.ala.org.au/archives/layers/sds-layers.tgz') }} dest="{{data_dir}}/biocache/layers/sds-layers.tgz" force=yes
-  when: sds_enabled is defined and sds_enabled == true
```
ansell: @sadeghim If you need this for appd, it may be simpler to add sds_enabled to the appd ansible file to to make all biocache ansible invocations download the sds-layers archive if they don't all need it

sadeghim: @ansell Actually it's needed even if the sds_enabled is False otherwise when you ingest a dataset it will throw exceptions. I think it's bug in biocache-store that it looks for the sds-layers while sds is not enabled.

PR #96, opened 2016-03-28
Update main.yml

PhillRob:


ansell:
Could you expand on the reasoning behind this change? We would need to update several other files to keep their tomcat memory allocation at 2G if we accept this change and it isn't immediately obvious why it is necessary.

PR #102, opened 2016-04-12
Rationalise solr schema

charvolant:
See https://github.com/AtlasOfLivingAustralia/bie-index/issues/45


charvolant:
Now out of date.

PR #105, opened 2016-06-16
Bie test on AWS servers

charvolant:
Stuff for deploying the new BIE/biocache to the AWS servers.

Note that this is not perfect yet. The setup requires a root app and service app on the same tomcat virtual host that our ansible scripts are not quite up to yet. As a result, some stuff needs to be manually moved around after deploying.


charvolant:
Updated for new name indexes

charvolant:
And more name indexes

Issue #133, opened 2017-07-03
Ubuntu-16.04 has mysql-5.7 as minimum supported version


ansell:
Ubuntu-16.04 has mysql-5.7 as its minimum supported version so we need to update from 5.6 to 5.7.

Issue #132, opened 2017-07-02
tomcat-7 with java-8 fails to install because default jre is not present also


ansell:
The tomcat-7-java-8 profile fails to install because the default jre from the ubuntu repositories is not present in addition to the oracle jdk 8 from the webupd8 PPA. Need to also install openjdk-8-jdk to fulfill the default jre role so that tomcat can install successfully. After tomcat installs we explicitly set JAVA_HOME in /etc/default/tomcat7 to point to the oracle jdk so the openjdk is not used in practice after the second tomcat restart. The first tomcat restart is during the tomcat install, where it fails if openjdk is not present before hand, as we can't add JAVA_HOME to the /etc/default/tomcat7 file because it is not present before the install.

PR #134, opened 2017-07-03
Deploy collectory-test with mysql-5.7

### LINKED ISSUE #131: namematching index not deployed for ala profile
If ``nameindex_variant=ala``, the namematching index is not deployed, due to the following commit:

https://github.com/AtlasOfLivingAustralia/ala-install/commit/9e07c59585b4d68b0d33af2708c33218ea54a9cd#diff-6adb85a81d8fcd33dca132aab338da67

Investigating whether there are any downsides to putting it back in.

This is a cause for some of the ansible jenkins failures

ansell:
Fixes #131 , #132 and #133 to get collectory-test deploying successfully to ubuntu 16.04. This does not include the collectory code changes that @charvolant is doing separately, it only includes ansible changes necessary to get the ansible scripts running again.

### DIFF
--- a/ansible/collectory-standalone.yml
+++ b/ansible/collectory-standalone.yml
@@ -1,8 +1,8 @@
 - hosts: collectory
   roles:
     - common
-    - java
-    - mysql
-    - tomcat
+    - java8
+    - mysql-5.7
+    - tomcat7-java8
     - apache
     - collectory
--- a/ansible/roles/java8/tasks/main.yml
+++ b/ansible/roles/java8/tasks/main.yml
@@ -2,6 +2,14 @@
   tags:
     - java8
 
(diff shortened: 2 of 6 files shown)


ansell:
@nickdos this PR includes the ansible role changes necessary to get the new Java-8 Tomcat apps installing on Ubuntu-16.04 and adds a ``mysql-5.7`` role for the apps that are updated to work with that version.

Basically, once someone in the sprint migrates an app, they need to switch its ansible file from using the ``java`` role to using the ``java8`` role, ``tomcat`` to ``tomcat7-java8`` and ``mysql`` to ``mysql-5.7`` to redepoly it on an Ubuntu.16.04 server.

PR #148, opened 2017-11-24
New config files for spatial-service and spatial-hub

ansell:
WIP for config files for new spatial-service/spatial-hub installation


ansell:
TODO: Add geonetwork

Issue #149, opened 2017-12-06
aws_name_indexes_repo files missing


reupost:
The aws_name_indexes_repo is set to http://s3.amazonaws.com/ala-nameindexes/latest in ansible/roles/common/tasks/setfacts.yml

Running the ala-demo.yml playbook fails on the following task:

TASK [nameindexer : Download Catalogue of Life 2012] ***********************************************************
fatal: [vagrant1]: FAILED! => {"changed": false, "dest": "/data/lucene/sources/dwca-col.zip", "msg": "Request failed", "response": "HTTP Error 404: Not Found", "state": "absent", "status_code": 404, "url": "http://s3.amazonaws.com/ala-nameindexes/latest/dwca-col.zip"}

http://s3.amazonaws.com/ala-nameindexes/latest seems to be missing the following resources that are referenced by nameindexer tasks:
dwca-col.zip
dwca-col-mammals.zip
IRMNG_DWC_HOMONYMS.zip
col_vernacular.txt.zip
(also merge_namematching.tgz but this may not be in use anywhere)

ansell:
Those files may have been added to S3 since you created the issue and the person who fixed it may have forgotten to comment here to let you know. They seem to exist for me now, with modification dates of 14 December 2017, a week after you commented here, giving the impression someone silently fixed the issue.

There are some issues with consistency (both size and the paths inside of the tar files) between the files deployed to ``archives.ala.org.au``/``ala-starr.it.csiro.au`` and those on ``s3.amazonaws.com`` but the person who created the name indexes isn't working this week so I can't resolve those until next week at the earliest.

Issue #137, opened 2017-07-13
biocache-cli role bad configuration


santiagomrv:
I get this error when i try to run ansible-playbook using the ale-demo.yml

	TASK [biocache-cli : unjar biocache] *******************************************
	fatal: [vagrant1]: FAILED! => {"failed": true, "msg": "file or module does not exist: /usr/lib/biocache/biocache.zip"}

I had to comment this lines in the /Users/gbif-apple/dev/als-install-andorra/ansible/roles/biocache-cli/tasks/main.yml file and unzip the file manually:

	#- name: unjar biocache
	#  unarchive:
	#    src: /usr/lib/biocache/biocache.zip
	#    dest: /usr/lib/biocache/
	#    remote_src: True
	#  tags:
	#    - biocache-cli

When i do this then the process finish correctly.

Santi!!!

santiagomrv:
Other think that i found it. Once I have the vagrant with all installed, I have realized that i have to change the configuration of /usr/bin/biocache because my virtual machine only has 6GB RAM and we set $JAVA_OPTS with that:

exec "$JAVACMD" $JAVA_OPTS -Xms2g -Xmx2g $BIOCACHE_OPTS -XX:+UseG1GC -Dfile.encoding=UTF8 -Dlog4j.configuration=log4j.xml -Dactors.corePoolSize=8 -Dactors.maxPoolSize=16 -Dactors.minPoolSize=8 -Djava.util.Arrays.useLegacyMergeSort=true

The result of that when we want to run $ biocache is this:

	OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00000003f6000000, 17179869184, 0) failed; error='Cannot allocate memory' (errno=12)
	#
	# There is insufficient memory for the Java Runtime Environment to continue.
	# Native memory allocation (malloc) failed to allocate 17179869184 bytes for committing reserved memory.
	# An error report file with more information is saved as:
	# /tmp/hs_err_pid13184.log
	vagrant@ala:/usr/lib/biocache/biocache-store-1.9$ sudo biocache
	OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00000003f6000000, 17179869184, 0) failed; error='Cannot allocate memory' (errno=12)

Maybe we could solve this detecting the RAM configuration of the virtual machine and set those values depending on it.

Thanks,
Santi!!

santiagomrv:
[...]

PR #150, opened 2017-12-12
Resource file paths

reupost:
These are the tweaks I needed to make to get the demo installation to run. There are definitely better ways of sorting out some of the paths (cf. issue #149), so I wouldn't necessarily accept these changes.

### DIFF
--- a/ansible/roles/biocache-cli/tasks/main.yml
+++ b/ansible/roles/biocache-cli/tasks/main.yml
@@ -51,6 +51,14 @@
   tags:
     - biocache-cli
 
+- name: reset biocache java memory default
+  replace:
+    dest: /usr/lib/biocache/biocache-store-{{ version }}/bin/biocache
+    regexp: '[-]Xm([s|x])\d\dg'
+    replace: '-Xm\g<1>1g'
+  tags:
+    - biocache-cli
+
 - name: make the symlink for /usr/bin/biocache
   file: src=/usr/lib/biocache/biocache-store-{{ version }}/bin/biocache dest=/usr/bin/biocache state=link
   tags:
--- a/ansible/roles/biocache-cli/vars/main.yml
+++ b/ansible/roles/biocache-cli/vars/main.yml
@@ -1,5 +1,5 @@
 # remote assets
-version: "{{ biocache_cli_version | default('1.9') }}"
+version: "{{ biocache_cli_version | default('1.9.2-SNAPSHOT') }}"
 artifactId: "biocache-store"
 classifier: "distribution"
 groupId: "{{ biocache_cli_group_id | default('au.org.ala') }}"
--- a/ansible/roles/biocache-cli/vars/main.yml
+++ b/ansible/roles/biocache-cli/vars/main.yml
@@ -1,5 +1,5 @@
 # remote assets
-version: "{{ biocache_cli_version | default('1.9.2') }}"
+version: "{{ biocache_cli_version | default('1.9') }}"
 artifactId: "biocache-store"
 classifier: "distribution"
 groupId: "{{ biocache_cli_group_id | default('au.org.ala') }}"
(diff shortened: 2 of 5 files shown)


ansible/roles/biocache-cli/vars/main.yml
```diff
@@ -1,5 +1,5 @@
 # remote assets
-version: "{{ biocache_cli_version | default('1.9.2') }}"
+version: "{{ biocache_cli_version | default('1.9') }}"
```
ansell: Could you backtrack these default changes as the baseline at ALA has already moved to these versions? You can override them in your local playbooks instead of changing the versions here.

reupost:
No problem, I've set the biocache-store to 1.9.2-SNAPSHOT

ansible/roles/biocache-cli/tasks/main.yml
```diff
@@ -51,6 +51,14 @@
   tags:
     - biocache-cli
 
+- name: reset biocache java memory default
```
ansell: This looks like it would cause every use of this ansible file to use 1GB as its maximum heap size (unless I am reading it wrong). 

You should be setting your preferred memory size in a ``biocache_opts`` variable in your ansible inventory file instead. That option will be used to override the defaults (which are already 1G) in ``ansible/roles/biocache-cli/templates/environmentSetupFile.sh``

reupost: Sorry, I know this was a bit hacky - happy to revert. It slipped my mind that this would find its way into the pull request.  It was a local fix for #137, and only affects the biocache configuration which is otherwise hardcoded as demanding 16GB RAM in the biocache.zip file, which seems a little high for an initial setup.  Changing the biocache_opts values in  the /ansible/inventories/vagrant/biocache-cli-vagrant file doesn't seem to affect this, by the way.

ansell: 16GB is required to run the ALA production copy. However, it isn't clear why it would default to that unless either you are manually creating the zip file using ``createProductionExec.sh`` or someone deployed a version to a maven repository using 16GB instead of the default 4GB. All versions deployed from Travis should get 4GB as the default value in their ``environmentSetupFile.sh``, but unusual things may happen if someone ran ``mvn deploy`` or ``mvn install`` locally with some unusual settings.

Does ``environmentSetupFile.sh`` exist on your system?

reupost: I have the environmentSetupFile.sh with the 1GB defaults in the biocache-cli/templates folder.  I didn't manually create the zip file.  It does seem like the biocache script hardcodes the memory and hence ignores any other configuration.  
The final command in the script on my ALA-demo server looks like:

`exec "$JAVACMD" $JAVA_OPTS -Xms1g -Xmx1g $BIOCACHE_OPTS -XX:+UseG1GC -Dfile.encoding=UTF8 -Dlog4j.configuration=log4j.xml -Dactors.corePoolSize=8 -Dactors.maxPoolSize=16 -Dactors.minPoolSize=8 -Djava.util.Arrays.useLegacyMergeSort=true -Dactors.maxPoolSize=8 \
  -classpath "$CLASSPATH" \
  -Dapp.name="biocache" \
  -Dapp.pid="$$" \
  -Dapp.repo="$REPO" \
  -Dapp.home="$BASEDIR" \
  -Dbasedir="$BASEDIR" \
  au.org.ala.biocache.cmd.CMD2 \
  "$@" `

(the -Xms1g -Xmx1g entries were 16g each before I replaced them, but should possible be omitted all together to allow the other memory configuration to take effect)

djtfmartin:
I have a feeling this is now out of date ? can i close or is there something still relevant here ? @ansell @reupost

ansell:
Yes, I think it is out of date. The ALA has been customising the memory usage for biocache-cli since this using an another method. The nameindex parts have also been changed since this.

reupost:
I'm happy for this to be closed. Thanks, guys.

PR #223, opened 2018-12-17
Fix few things in ala-install

JeGoi:
The first commit is fixing the missing config files for biocache-service.
The second commit is fixing the path for lucene if the nameindex_variant is not ala
Thank you


JeGoi:
Sorry few things are missing

PR #224, opened 2018-12-17
Fix few things in ala-install

JeGoi:
The first and third commits are fixing the missing config files for biocache-service.
The second commit is fixing the path for lucene if the nameindex_variant is not ala.
The 4th is updating the solr7 link and version the version 7.4.0 is no more available on spango.com
Thank you


JeGoi:
Hum something is wrong !

Issue #232, opened 2019-01-29
Support nginx max_conns parameter where possible


ansell:
Starting from nginx `1.11.5` the `max_conns` parameter is able to be used to limit the number of connections from nginx to `proxy_pass` locations. This would be advantageous for avoiding tomcat thread spikes when requests spike, as happened last night and interrupted access to biocache downloads.

`https://nginx.org/en/docs/http/ngx_http_upstream_module.html`

Ubuntu-16.04 seems to offer `1.10.3` by default, so we would need to switch to the official NGINX PPA to use it there.

Ubuntu-18.04 seems to offer `1.14.0`, so it can be used there without any changes to the apt repositories.

ansell:
Stress testing on biocache service shows that small values for `max_conns` can be viable for low load servers. There was a 15% error rate on a 10 minute stress test for `biocache-service` on `test-bws-b5-2` with `max_conns=8`. A repeat test (without clearing the nginx cache or restarting tomcat) reduced the error rate to 8%.

PR #234, opened 2019-01-30
issue #232 : Initial support for max_conns in nginx configuration

ansell:
If there is at least one item in the `nginx_paths` list with `is_proxy=True`, this will create an `upstream` unique to the `hostname` and the proxied IP (currently hardcoded to `127.0.0.1` which matches the rest of the configuration) and the proxied port.

By default it allows 50 concurrent connections to the proxied server, but this is configurable. Testing for this was done on `test-bhub-b5-2` and `test-bws-b5-2`, with `nginx_max_conns` set to `8` to verify that it blocks on multiple queries.

No stress testing was performed as those instances are connected to the production solr cloud.

The `nginx_load_balancing` setting must be switched on to opt-in to the new features so they won't influence current servers that are not trialing this. Eventually, it may be switched on by default.

Signed-off-by: Peter Ansell <p_ansell@yahoo.com>

### DIFF
--- a/ansible/roles/nginx_vhost/defaults/main.yml
+++ b/ansible/roles/nginx_vhost/defaults/main.yml
@@ -49,3 +49,8 @@ nginx_cache_valid_time: "60m"
 nginx_cache_invalid_time: "0s"
 # Set this to true in your inventory if you know that you are only deploying a single nginx_vhost and you want to clear previous versions before deploying each time
 clear_vhost_fragments: false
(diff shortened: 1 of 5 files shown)


ansell:
Note also that this requires at least nginx-1.11.5. This requires the official nginx ppa when installing on Ubuntu-16.04 (which is now the default, but may have been manually switched to false in some user setups)

PR #242, opened 2019-03-11
roles: biocache-service, nameindex update

JeGoi:
Change biocache-service config files installation from files to template
Remove hard coded ALA from config files

### DIFF
--- a/ansible/roles/solr7/tasks/main.yml
+++ b/ansible/roles/solr7/tasks/main.yml
@@ -125,7 +125,7 @@
     - solr7_create_cores
 
 - name: Create SOLR cores
-  get_url: url="{{ solr_base_url }}/solr/admin/cores?action=CREATE&name={{ item }}&instanceDir=/data/solr/data/{{ item }}&config=solrconfig.xml&dataDir=data" force=yes timeout=30 dest=/tmp/
+  get_url: url="{{ solr_base_url }}/solr/admin/cores?action=CREATE&name={{ item }}&instanceDir={{data_dir}}/solr/data/{{ item }}&config=solrconfig.xml&dataDir=data" force=yes timeout=30 dest=/tmp/
   with_items:
     - "biocache"
     - "bie"


JeGoi:
Add a fix for name matching

ansible/roles/biocache-service/tasks/main.yml
```diff
@@ -68,36 +68,6 @@
         use_cache: true
         force_cache: true
         proxy_pass: "http://127.0.0.1:8080{{biocache_service_proxy_target}}/occurrence/legend"
-      - path: "{{biocache_service_context_path}}/occurrence/facets"
```
ansell: These removals need to be replaced with a configurable approach. The ALA rely heavily on these rules to cater for the amount of traffic that we serve on our low budget hardware.

ansible/roles/nameindex/tasks/main.yml
```diff
@@ -48,21 +48,20 @@
   tags:
     - "nameindex"
 
-## Ensure that data_dir/lucene exists as a directory
-
+## Ensure that data_dir/lucene exists as a directory (ALA and not-ALA)
```
ansell: Can you split the nameindex changes out into a separate pull request please?

ansible/roles/biocache-service/tasks/main.yml
JeGoi: Right, I've miss them. They are not part of my modification.
I'll be back with an update.

ansible/roles/nameindex/tasks/main.yml
JeGoi: Right, I'll do it

djtfmartin: thanks @JeGoi

JeGoi:
I've created two new pull request for each and with the missing lines.
Sorry, for that unfortunate deletion and happy that @ansell have seen that.

Issue #253, opened 2019-04-01
Duplicate specieslist entries in ngnix


vjrj:
Following the conversation with @ansell in https://github.com/AtlasOfLivingAustralia/ala-install/commit/ef68ea0379ccdf369548934af7eec62c8f0f5375

I get a duplicate entry of `specieslist` re-runnging the workshop inventories with:
```
ansible-playbook --private-key somekey -u ubuntu -s -i ansible/inventories/workshop/demo-livingatlas.yml ansible/ala-demo.yml --skip "nameindex,solr7" -vvvv
```

Error:

``` 
# systemctl status nginx.service 
(...)
Apr 01 06:27:23 demo.livingatlas.org nginx[24185]: nginx: [emerg] duplicate location "/specieslists" in /etc/nginx/sites-enabled/demo.living
Apr 01 06:27:23 demo.livingatlas.org systemd[1]: nginx.service: Control process exited, code=exited status=1
```
also:
```
# grep location '/etc/nginx/sites-enabled/demo.livingatlas.org.conf' 
    (...)
    location /specieslists {
    location /biocache-download {
    location /ala-bie/geo {
    location /bie-index {
    location /ala-bie {
    location /biocache-service {
    location /ala-hub {
    location /collectory {
    location /logger-service {
    location /specieslists {
```

vjrj:
@ansell there is a `bie-index` tag in `roles/species-list/tasks/main.yml` that maybe is the cause...(I'll test it)

ansell:
Thanks, it didn't occur to me that the demo may have imported it twice. I was running it on dedicated instances of specieslists when testing it.

PR #254, opened 2019-04-01
issue #253 : Test clearing vhost fragments

ansell:
Test clearing vhost fragments for demo.livingatlas.org

It is possible that the cause of issue #253 was a change in the name of the fragment file. The only known fix for this is to clear the vhost fragments completely at the start of the ansible run.

IMPORTANT: This unfortunately breaks the ability to use tags to skip any `nginx_vhost` role calls as doing so will result in an incomplete nginx configuration.

CC @vjrj

### DIFF
--- a/ansible/inventories/workshop/demo-livingatlas.yml
+++ b/ansible/inventories/workshop/demo-livingatlas.yml
@@ -294,3 +294,6 @@ postgis_version=9.6
 pg_service=postgresql-9.6
 pg_version=9.6
 postgis_version=2.4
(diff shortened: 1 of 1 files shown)


vjrj:
Works for me and fix #253, thanks indeed @ansell.

PS: Maybe to split the nginx configuration and to use `include` is another option in the future:
https://serverfault.com/questions/707955/nginx-split-large-configuration-file

djtfmartin:
thanks a stack @ansell !

PR #268, opened 2019-04-30
Added login/logout CAS urls to collectory config

vjrj:
Added some CAS login/logout urls to collectory config so its not redirect to `auth.ala.org.au` in non-ALA nodes using CAS.


ansible/roles/collectory/templates/config/collectory-config.properties
```diff
@@ -9,6 +9,11 @@ security.cas.authenticateOnlyIfLoggedInFilterPattern={{ collectory_authenticate_
 security.cas.uriExclusionFilterPattern={{ collectory_exclusion_filter_pattern | default('/images.*,/css.*,/js.*') }}
 security.cas.appServerName={{ collectory_base_url }}
 security.cas.casServerUrlPrefix={{ auth_base_url }}/cas
+security.cas.casServerLoginUrl={{ auth_base_url }}/cas/login
```
sbearcsiro: These settings (`security.cas.casServerLoginUrl` and `security.cas.casServerLogoutUrl`) aren't used in the auth plugin, so shouldn't be necessary unless the app itself uses them?

vjrj: Maybe we have to patch:
https://github.com/AtlasOfLivingAustralia/ala-collectory/blob/master/grails-app/conf/Config.groovy
instead?

PR #276, opened 2019-05-06
Update download url in nginx

JeGoi:
If you are not using a subdomain like `biocache-ws.hello.com` but you are using a context-path like `data.hello.com/biocache-ws`, on line 229 "/biocache-download" will failed


ansell:
Thanks, looks good to me.

PR #288, opened 2019-06-03
CA5-5 living-atlas demo inventory

vjrj:
An inventory to install CAS-5 in a living-atlas demo.

I modified `ansible/aws-cas-5.yml` to make it work with this inventory adding new simple mongodb-org install role, and moving code to that role.

This works great in new CAS-5 simple installations but maybe not so good in current CAS-5 multi-machine deployments. So maybe, a new cas-5 demo playbook is better that to modify `aws-cas-5.yml`, I don't know. I was trying to not duplicate code.


vjrj:
I've just added also to the LA yeoman generator, so it's more easy to test:
https://github.com/vjrj/generator-living-atlas

djtfmartin:
Muchas gracias Vicente !

vjrj:
Un placer!

JeGoi:
May I ask why you finally did not chose to create(/rename) a new playbook for cas5 ?
With `aws-cas-5.yml`, it really looks like it's only available for AWS, doesn't it?
Regards

vjrj:
Hi @JeGoi, nice to read you ;-)  

Wel'll in fact I did not renamed for three reasons:
- To try not to break things to those using this playbook (for instance in a continuous integration system)
- To show what I changed step by step to make this work in small steps (this is the last of other cas-5 related PRs).
- in beyond my responsibilities to refactor/reorganize/rename these files/directories but... I would prefer other name, yes

In general I was trying to facilitate the merge... so, now is up to the ALA developers to rename, etc.

PR #290, opened 2019-06-05
Make external applications secret/keys optional

vjrj:
A small improvement to activate external auth in CAS-5 only when valid secret/keys are defined.

These prevent to touch `application.yml` if you use other external auths (like github, etc) in your LA node different than `.au`.


djtfmartin:
gracias @vjrj :)

vjrj:
Gracias a ti @djtfmartin ...

PR #295, opened 2019-06-11
Setting a pair of def vars

vjrj:
Just setting a pair of two new default vars.

### DIFF
--- a/ansible/roles/bie-index/templates/bie-index-config.yml
+++ b/ansible/roles/bie-index/templates/bie-index-config.yml
@@ -31,7 +31,7 @@ layers:
 images:
   # URL of image configuration file
   config: file://{{ data_dir }}/bie-index/config/image-lists.json
-  base: https://images.ala.org.au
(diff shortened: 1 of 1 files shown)


ansell:
Thanks, will tweak the ALA URL variable to match the rest by changing it to `ala_base_url` after I merge.

vjrj:
Thanks @ansell !

vjrj:
HI @ansell, just thinking in the renaming.

As some portals like `.es` or `.fr` (if I'm not wrong) use different urls for main site and for ALA software, see:
- https://www.gbif.es -> wordpress 
- https://datos.gbif.es -> this is a static landing page, and skin header/footer provider

or
- http://www.gbif.fr/
- http://portail.gbif.fr/

and this is used for get the sitemap.xml, maybe it's better to maintain `wordpress_service_url` variable with a default to `ala_base_url` variable, so only sites like `.es` will use in their inventories.

Thanks!

PR #301, opened 2019-06-20
Added collectory uriFilterPattern allowing to update ipt providers when using CAS

vjrj:
Added `ws/ipt/scan` to `uriFilterPattern` allowing to update `ipt` providers when using CAS in `collectory`.


vjrj:
Without this, you get a `Unable to create resources for dp2`
Accesing to `ws/ipt/scan/dp2.json?create=true`

PR #305, opened 2019-07-09
Fix postgis var definition checks

vjrj:
After https://github.com/AtlasOfLivingAustralia/ala-install/commit/f079973be51fe2c198c89d67eaa4bcbac2526ccf new `postgis` var check fails with:

```
fatal: [core.biodiversityatlas.at]: FAILED! => {"msg": "The conditional check 'postgis_version is defined and postgis_version | length > 0' failed. The error was: Unexpected templating type error occurred on ({% if postgis_version is defined and postgis_version | length > 0 %} True {% else %} False {% endif %}): object of type 'float' has no len()\n\nThe error appears to be in '/home/ubuntu/Desktop/bio.at/ala-install/ansible/roles/postgresql/tasks/postgis.yml': line 1, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: install setfacl support\n  ^ here\n"}
```

or in summary, with `'float' has no len()` when using `postgis_version = 2.4`.

I just follow `ansiblelint` [suggestions](https://github.com/ansible/ansible-lint/blob/master/lib/ansiblelint/rules/ComparisonToEmptyStringRule.py) but please double check.

cc @ansell

### DIFF
--- a/ansible/roles/postgresql/tasks/main.yml
+++ b/ansible/roles/postgresql/tasks/main.yml
@@ -25,14 +25,13 @@
     - info
 
 - include: ./postgis.yml
-  when: postgis_version is defined and postgis_version | length > 0
+  when: postgis_version
   tags:
     - postgresql
     - postgis
 
 - include: ./gdal.yml
-  when: postgis_version is defined and postgis_version | length > 0
+  when: postgis_version
   tags:
     - postgresql
     - postgis
-
--- a/ansible/roles/postgresql/tasks/postgis.yml
+++ b/ansible/roles/postgresql/tasks/postgis.yml
@@ -2,7 +2,7 @@
   become: yes
   apt:
     pkg: acl
-  when: postgis_version is defined and postgis_version | length > 0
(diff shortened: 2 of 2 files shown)


ansell:
Try using `postgis_version="2.4"`

vjrj:
I imagine that should work but I din't try before because many component versions are without quotes.

Do you suggest that I should send a PR with the quotes?, for instance here:
```
$ grep -r "postgis_version=" ansible
ansible/inventories/vagrant/demo-vagrant:postgis_version=9.6
ansible/inventories/vagrant/demo-vagrant:postgis_version=2.4
ansible/inventories/workshop/demo-livingatlas.yml:postgis_version=2.4
```

vjrj:
Probably, can be deleted: 
```
ansible/spatial.yml:    - {role: postgresql, pg_service: 'postgresql-9.6', pg_version: '9.6', postgis_version: '2.4'}
```

PR #316, opened 2019-07-29
Include again common and java role in CAS

vjrj:
After installing CAS in a new server CAS service don't start, I see that is because java is not installed:
```
root@auth:/var/log/atlas# /opt/atlas/cas/cas.jar
Unable to find Java
```
and this was removed from the last CAS roles refactorization.


vjrj:
Thanks @djtfmartin !

PR #327, opened 2019-08-28
add download.shp.enabled to biocache-service config

adam-collins:


ansell:
Fits the new property defined in: https://github.com/AtlasOfLivingAustralia/biocache-service/pull/398/files#diff-ae43ce90f80f57f697dfc83a872e816dR243

Merging so that the test system will have this property available for testing.

PR #337, opened 2019-10-02
Set images port in demo playbook and service configuration

vjrj:
To not conflict with tomcat 8080 port we set some default port for image service in the demo playbook but also in the service configuration (so embedded tomcat not uses 8080 by default).

PS: I updated also the generator  to use the new image service:
https://github.com/vjrj/generator-living-atlas/commit/2ac96bd68498e739c8b9c292d03f48166713e102


ansell:
@vjrj Can you add a new `image-service/defaults/main.yml` file and put `8080` as the default value in for it.

vjrj:
@ansell done!

ansell:
Thanks, I will test this and merge it

vjrj:
Thanks @ansell

PR #339, opened 2019-10-03
Update webapi role

vjrj:
I've updated the webapi role to install https://api.gbif.es
I updated also the ansible generator to support webapi service, so other portals can use it also: https://github.com/vjrj/generator-living-atlas/commit/788a8698dcfba41abd2950d708915e6f84a4047c

So I update that role with the problems I found (CAS issues, old mysql version, old war used, etc).


ansell:
@Rita-C is working on the webapi code currently and would be the best person to review this.

PR #353, opened 2019-11-12
Improvements to spatial-hub role

vjrj:
Some improvements to the spatial-hub role, like:
- added the api key var to communicate with spatial-service 
- default area configurable
- private directory creation to avoid 
```
IntersectConfig [ERROR] Config error. Property "ANALYSIS_TMP_LAYER_FILES_PATH" with value "/data/spatial-data/private"  is not a valid local file path.  It does not exist.
```
I tested the default area, with default values (so Australia values are the same like without this patch) and other non Australia areas. See it in:
https://espacial.gbif.es


djtfmartin:
gracias @vjrj !

vjrj:
Thanks for your fast review @djtfmartin

Issue #356, opened 2019-11-14
create images databse


kantobogy:
After execute
#> ansible-playbook -i inventories/vagrant/demo-vagrant ala-demo.yml --private-key ~/.vagrant.d/insecure_private_key --user vagrant --become

the fatal error is:
fatal: [vagrant1]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'image_service_db_name' is undefined\n\nThe error appears to be in '/home/leonardo/Downloads/ala-install/ansible/roles/image-service/tasks/main.yml': line 15, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: create images database\n  ^ here\n"}

ansell:
@kantobogy I patched the variable names which had changed recently, can you test again after pulling the latest commit?

vjrj:
I'm doing the same with `ansible/inventories/living-atlas` but didn't have time yesterday to pull. In some minutes...

ansell:
This should be fixed, reopen if it is still an issue.

Issue #357, opened 2019-11-15
AnsibleUndefinedVariable (bie-hub- speciesList_api_key)


kantobogy:
Dear all

After execute
#> ansible-playbook -i inventories/vagrant/demo-vagrant ala-demo.yml --private-key ~/.vagrant.d/insecure_private_key --user vagrant --become

TASK [bie-hub : copy all config.properties] *******************************************************************************************************************************
fatal: [vagrant1]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'speciesList_api_key' is undefined"}

Thanks in advanced

vjrj:
Hi @kantobogy I just send a PR with some updated inventories. 

This var was added some days ago. you can simple put some value:
https://github.com/AtlasOfLivingAustralia/ala-install/pull/358/files#diff-a2e1c7737b7f7677e82c8068f08acb2eR273

vjrj:
I also updated the generator similarly:
https://github.com/living-atlases/generator-living-atlas/commit/3bca30df7dae7632ef9f1bb6afa31dc6ce3999ed

louiepang:
hi All,

I still have the fatal exception even checked the speciesList_api_key existed.
I tried to export from environment before run the  ansible-playbook but the result is same.

Any further idea that I can verify for my installation?

vjrj:
Hi @louiepang see this inventory sample:
https://github.com/AtlasOfLivingAustralia/ala-install/blob/master/ansible/inventories/living-atlas#L273
did you added the var like this?

If you problem persist, do you have access to our slack channel to contact me via DM so I try to help you there?

vjrj:
I tested several times ala-install without this problem.

PR #358, opened 2019-11-15
Living altas demo inventories updated. Added spatial inventory

vjrj:
I updated a little bit the LA inventories and added also the spatial one (using the generator and the Paris workshop inventories as reference).

Related with:
https://github.com/AtlasOfLivingAustralia/ala-install/issues/356


vjrj:
Also #357

PR #348, opened 2019-10-30
Update image-service.yml

meshack007:

PR #361, opened 2019-12-03
missing space prevents ansible to run

rpfigueira:


vjrj:
Wow. 

There is an already pending PR from Tanzania team: https://github.com/AtlasOfLivingAustralia/ala-install/pull/348 but as I didn't notice, I was asking for more info or a better description.

PR #365, opened 2020-01-23
Added missing default values for mdba list

vjrj:
Without default values we have some errors like:
``` 
fatal: [records.no.l-a.site]: FAILED! => {“changed”: false, “msg”: “AnsibleUndefinedVariable: ‘specieslist_uid’ is undefined”}
``` 
PS: Thanks to Rukaya from gbif.no for the feedback.


vjrj:
> The line here should be wrapped in an `if` statement

Done and tested, thanks @ansell

ansell:
LGTM

Issue #256, opened 2019-04-02
Rethink the nginx vhost fragments system


ansell:
The biggest issue right now for nginx server/location configurations is the requirement to be able to deploy to the same domain name from different playbooks at different times.

The second biggest issue is that we are not declaring the domain name(s) in the original call to the `webserver` role (which calls the `nginx` role), only in calls to the `nginx_vhost` role. This could be improved, but there is no urgency at this point so it is fairly low priority. 

The third biggest issue is the sporadic use of `tag` to isolate parts of a playbook for partial use, in which case we can't guarantee that a previous element will always be called. Novice users of tags may think the system is broken because we appear to be supporting tags even though they are always going to be broken in strange and unusual ways if used by non-experts due the complex dependencies between roles due to the way `nginx_vhost` is pulled in by applications and the domaon .

Some users also merge multiple applications that could easily be supported in separate domain names into single playbooks which make them more cumbersome to support than they should be. Unfortunately there is no technical direction at this point to fix those issues so they are also a complicating factor in being able to manage the nginx configuraiton at a higher level than the `nginx_vhost`+`vhost_fragments` persistence level.

In the multiple playbooks case, we can't touch any of the previous nginx configuration for fear of removing an essential element, or duplicating an essential element.

In the `nginx_vhost` case, it is too late to be able to clean up previous fragments automatically at that stage, as a previous call to `nginx_vhost` may have already modified current fragments.

Using `include` to pull in separate parts of the configuration would make it cleaner, but it wouldn't fix the requirements that we need to allow for in this case.

The current requirement is for users who know they are deploying from a single playbook to setup the vhost fragments they want to clear in their inventory file, which is picked up by the high level `nginx` role:

```
nginx_vhost_fragments_to_clear=["demo.livingatlas.org"]
```

This step is optional as it can't be implemented any other way at this point without users having other complaints which we would then need to deal with individually to verify the cause for them.

This issue is open for suggestions on how to improve the overall nginx setup while keeping to the current (and likely future) requirements of different playbooks, and partial executions.

cc @vjrj

vjrj:
I'm my experience using other `nginx` roles like:
https://github.com/nginxinc/ansible-role-nginx
or the `puppet` `nginx` module they allow incremental adding of vhost but they don't care about removing vhosts. 

So they create a vhost file in `/etc/nginx/sites-*` when you configure a new vhost proxy or web server.

[...]

PR #366, opened 2020-01-27
Prevent overwrite vhost configurations when deploying in same server

vjrj:
If you deploy `https://biocache.l-a.site/` and later `https://biocache.l-a.site/ws/` in the same server, at the end `ningx_vhost` ansible role will only configure the last one (because the first is overwritten).

This patch solved this adding a new ansible variable `ws_same_host` setted to false for backward compatibility and generating two nginx vhost like `biocache.l-a.site-ws.conf` and `biocache.l-a.site-ws.conf`.

PS: Thanks to @rukayaj from gbif.no for re-testing this patch.

### DIFF
--- a/ansible/aws-cas-5.yml
+++ b/ansible/aws-cas-5.yml
@@ -66,6 +66,7 @@
       login_port: "{{ item.login_port }}"
       login_user: "{{ item.login_user }}"
       login_password: "{{ item.login_password }}"
+      encoding: "{{ cas_db_encoding | default('utf8') }}"
     run_once: true
     with_items:
     - { name: "{{ cas_db_name | default('emmet') }}", login_host: "{{ cas_db_hostname }}", login_port: "{{ cas_db_port | default('3306') }}", login_user: "{{ mysql_root_username }}", login_password: "{{ mysql_root_password }}" }
--- a/ansible/collectory-standalone.yml
+++ b/ansible/collectory-standalone.yml
@@ -5,5 +5,6 @@
     - webserver
     - mysql-5.7
     - tomcat
+    - i18n
     - collectory
   become: yes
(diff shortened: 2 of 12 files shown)


ansell:
I don't think this is a viable approach in the long term, however, I am no longer allocated to working on ansible playbooks on the ALA end of things.

Is it possible to use the `appname` variable that I added as the first commit for #256 rather than using the context path to create a suffix? https://github.com/AtlasOfLivingAustralia/ala-install/commit/7ad7af47ae5e28b2af964b77a4ba922bc04bc802

vjrj:
My opinion is for a better resolution we should use includes instead of fragments and merge, as other nginx and apache roles do and as described in https://github.com/AtlasOfLivingAustralia/ala-install/issues/256#issuecomment-515159205

I tried to do a PR before with this approach but I didn't finished it, and the changes to the `nginx_vhost` role were more complex (and difficult to accept) than this little one. 

I added an entry here to explain that because is a recurrent issue with every new deployment:
https://github.com/AtlasOfLivingAustralia/documentation/wiki/FAQ#should-i-use-subdomains-or-pathnames-for-my-la-services

This PR don't affect current deployments or inventories, because with the default value will generate the same nginx-configuration.

About use `appname` or `context` SGTM to use `appname`. To summarize, nowadays, without this PR (or setting the new var):

- `https://biocache.l-a.site` ->  `biocache.l-a.site.conf`
- `https://biocache.l-a.site/ws` ->  `biocache.l-a.site.conf` (overwritten)

using `context` and setting the new var: 

- `https://biocache.l-a.site` ->  `biocache.l-a.site.conf`
- `https://biocache.l-a.site/ws` ->  `biocache.l-a.site-ws.conf` (new one)

with `appname` and setting the new var:

- `https://biocache.l-a.site` ->  `biocache.l-a.site-biocache_hub-service.conf` (new one)
- `https://biocache.l-a.site/ws` ->  `biocache.l-a.site-biocache-service.conf` (new one)

So I'll modify this PR with this last option.

Thanks @ansell .

vjrj:
Sorry I messed this PR merging with upstream.

I'll send a new one.

PR #368, opened 2020-02-11
Prevent overwrite vhost configurations when deploying in same server (rewritten with appname)

vjrj:
I rewrote this https://github.com/AtlasOfLivingAustralia/ala-install/pull/366 following @ansell suggestion.

I tested in a demo site and with `true` in the new var `vhost_with_appname_conf` and I get:

```
biocache.l-a.site-ala-hub.conf
biocache.l-a.site-biocache-service.conf
```
as expected.

### DIFF
--- a/ansible/roles/nginx_vhost/tasks/main.yml
+++ b/ansible/roles/nginx_vhost/tasks/main.yml
@@ -410,3 +410,9 @@
   tags:
     - nginx_vhost
 
+- name: unset vfragments_suffix
+  set_fact:
+    vfragments_suffix: ''
+  tags:
+    - nginx_vhost
+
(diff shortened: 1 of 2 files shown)


ansible/roles/nginx_vhost/tasks/main.yml
```diff
@@ -62,6 +62,20 @@
   tags:
     - nginx_vhost
 
+- name: set vfragments_suffix when vhost_with_appname_conf is true
```
ansell: Just for avoid this fact leaking out of this role, can you also `unset` this variable at the end of this file, similar to: ```
- name: unset nginx_paths
  set_fact:
    nginx_paths: ''
  tags:
    - nginx_vhost
```

vjrj: Added and tested, thanks for the feedback

PR #370, opened 2020-02-13
Added CORS fragments to nginx_vhost role

### LINKED ISSUE #225: Add CORS header support to nginx locations
NGINX `location`s need to have CORS support optionally available for applications to specify origins/verbs/etc. that are allowed to perform cross-site queries. This is being done on an ad hoc manner on some servers currently, which makes it impossible to deploy using Ansible to those servers as it would break core functionality that has been manually added.

If `add_header` is used inside an NGINX `location`, the other security headers (HSTS/Referrer-Policy/etc.) which are defined in the enclosing `server` also need to be reinserted into the `location` or they are ignored.

vjrj:
Following [this conversation in slack](https://atlaslivingaustralia.slack.com/archives/CCTFGEU1G/p1581527510131800) this fix #225.

Tested in a demo site.

### DIFF
--- a/ansible/roles/nginx_vhost/tasks/main.yml
+++ b/ansible/roles/nginx_vhost/tasks/main.yml
@@ -146,6 +146,14 @@
   tags:
     - nginx_vhost
 
+- name: copy cors configuration if required
+  template:
+    src: "ala_cors.j2"
+    dest: "{{nginx_conf_dir}}/conf.d/ala_cors"
+  when: nginx_cors_origin_regexp is defined
+  tags:
+    - nginx_vhost
+
 - name: add upstream fragment
   template:
     src: "fragment_02_upstream.j2"
@@ -273,6 +281,16 @@
   tags:
     - nginx_vhost
 
+- name: add cors fragment if required
+  template:
+    src: "fragment_74_location_cors.j2"
+    dest: "{{nginx_conf_dir}}/vhost_fragments/{{hostname}}/http_70_location_{{ item.sort_label | default(item.path | basename) }}_74_cors"
+  with_items:
+    - "{{ nginx_paths}} "
+  when: (ssl | bool == False or (ssl | bool == True and force_https | bool == False)) and vhost_required | bool == True and nginx_cors_origin_regexp is defined
+  tags:
+    - nginx_vhost
+
 # if not ssl or (ssl and not force_https): copy (75 per path) http
 # basename filter returns last part of path e.g. for /ws returns ws, for / returns empty
 - name: if not using force_https, add location http fragments
@@ -351,6 +369,16 @@
   with_items:
     - "{{ nginx_paths }}"
   when: ssl | bool == True and vhost_required | bool == True
+  tags:
+    - nginx_vhost
+
+- name: add cors fragment if required
+  template:
+    src: "fragment_74_location_cors.j2"
+    dest: "{{nginx_conf_dir}}/vhost_fragments/{{hostname}}/https_70_location_{{ item.sort_label | default(item.path | basename) }}_74_cors"
+  with_items:
+    - "{{ nginx_paths}} "
+  when: ssl | bool == True and vhost_required | bool == True and nginx_cors_origin_regexp is defined
   tags:
     - nginx_vhost
 
--- a/ansible/roles/nginx_vhost/tasks/main.yml
+++ b/ansible/roles/nginx_vhost/tasks/main.yml
@@ -149,7 +149,7 @@
 - name: copy cors configuration if required
   template:
     src: "ala_cors.j2"
-    dest: "{{nginx_conf_dir}}/conf.d/ala_cors_{{appname}"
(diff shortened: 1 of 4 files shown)


ansible/roles/nginx_vhost/templates/ala_cors.j2
```diff
@@ -0,0 +1,57 @@
...
+
+# This is modified version of: https://enable-cors.org/server_nginx.html restricting to previous regexp
+
+if ($cors = COPTIONS) {
+  add_header 'Access-Control-Allow-Origin' '*';
+  add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
+  #
+  # Custom headers and headers various browsers *should* be OK with but aren't
+  #
+  add_header 'Access-Control-Allow-Headers' 'DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range';
```
ansell: Can you setup an array variable for the contents of these headers (also including `Access-Control-Allow-Origin` and `Access-Control-Allow-Method` so we can tune them per application as necessary?

ansible/roles/nginx_vhost/tasks/main.yml
```diff
@@ -273,6 +281,16 @@
   tags:
     - nginx_vhost
 
+- name: add cors fragment if required
+  template:
+    src: "fragment_74_location_cors.j2"
+    dest: "{{nginx_conf_dir}}/vhost_fragments/{{hostname}}/http_70_location_{{ item.sort_label | default(item.path | basename) }}_74_cors"
```
ansell: The destination is named `http_70_location`. I think that should be `http_74_location` to ensure the fragment numbering matches the template numbering for debugging purposes.

ansible/roles/nginx_vhost/tasks/main.yml
```diff
@@ -354,6 +372,16 @@
   tags:
     - nginx_vhost
 
+- name: add cors fragment if required
+  template:
+    src: "fragment_74_location_cors.j2"
+    dest: "{{nginx_conf_dir}}/vhost_fragments/{{hostname}}/https_70_location_{{ item.sort_label | default(item.path | basename) }}_74_cors"
```
ansell: The destination is named `http_70_location`. I think that should be `http_74_location` to ensure the fragment numbering matches the template numbering for debugging purposes.

ansible/roles/nginx_vhost/tasks/main.yml
```diff
@@ -146,6 +146,14 @@
   tags:
     - nginx_vhost
 
+- name: copy cors configuration if required
+  template:
+    src: "ala_cors.j2"
+    dest: "{{nginx_conf_dir}}/conf.d/ala_cors"
```
ansell: This needs to be named including the `appname` so that we can tune it as necessary for each application and not have it interfere with other applications.

ansible/roles/nginx_vhost/tasks/main.yml
vjrj: the destination is `https_70_location_1_74_cors` so this is the result:
```
https_70_location_1_70_start
https_70_location_1_73_content
https_70_location_1_74_cors
https_70_location_1_75_end
```

ansible/roles/nginx_vhost/templates/ala_cors.j2
vjrj: I'll do

ansible/roles/nginx_vhost/tasks/main.yml
vjrj: See https similar comment below.

ansible/roles/nginx_vhost/tasks/main.yml
vjrj: I'll do, thanks.

ansible/roles/nginx_vhost/templates/ala_cors.j2
```diff
@@ -0,0 +1,57 @@
...
+}
+
+if ($request_method = OPTIONS) {
+  set $cors '${cors}OPTIONS';
+}
+
+# This is modified version of: https://enable-cors.org/server_nginx.html restricting to previous regexp
+
+if ($cors = COPTIONS) {
+  add_header 'Access-Control-Allow-Origin' '*';
```
vjrj: I think is not necessary to change the origin header here because is already done in line 10 via a regexp, so this code isi only accessed to that domains

vjrj:
Since I pushed this PR we have manually applied (*) and tested in several LA deployments (Tanzania, Guatemala, Royal Gardens Kew demo, Vermont). I think is safe to merge it if you don't have any other comment.

Thanks,

(*) With people not very comfortable with git, this is quite a difficult task. Yesterday I spend more than an hour (looking my chat, from 15:41 to 17:35) just trying to help someone in slack, to apply this PR, to solve the CORS errors in a LA site. And it's not the first time this happens to me.

vjrj:
Merged after talking with @ansell , thanks

PR #373, opened 2020-02-26
Added missing appname in image-service

vjrj:
After https://github.com/AtlasOfLivingAustralia/ala-install/pull/368 we need to set missing `appname` in `image-service` as in the rest of roles, to prevent these kind of errors:

```
fatal: [imagenes.snib.conap.gob.gt]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'appname' is undefined\n\nThe error appears to be in '/home/gt/ala-install/ansible/roles/nginx_vhost/tasks/main.yml'
```

### DIFF
--- a/ansible/roles/image-service/tasks/main.yml
+++ b/ansible/roles/image-service/tasks/main.yml
@@ -55,6 +55,7 @@
   include_role:
     name: nginx_vhost
   vars:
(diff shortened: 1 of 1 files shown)


ansell:
LGTM

PR #385, opened 2020-04-24
Prevent cassandra role to rm -rf its data in playbook reruns

vjrj:
This week in Vermont, when they were updating their LA deployment to new LA versions, the cassandra role wiped all its data after rerun it.
https://github.com/AtlasOfLivingAustralia/ala-install/blob/d610dceee1b85ea494fa66c83af63d0025fdc23a/ansible/roles/cassandra3/tasks/main.yml#L103

This patch tries to prevent this to happen again but also to rerun the cassandra role safely (for upgrades).

Take into account that many LA deployments right now doesn't have a cassandra or solr cluster.

To detect if the cassandra remove task should be executed or not, I use a simple initial detection of the cassandra package installation status. 

If the cassandra package is the fist time is installed, the rm task is executed at the end:

2020-10-24-22:27:23-screenshot-cassandra-rm

but if the cassandra package is already installed, this rm tasks is skipped:

2020-10-24-22:27:23-screenshot-cassandra-not-rm

cc @jloomisVCE 

PS: I added a similar detection for RedHat but the rm task is only used in Debian

### DIFF
--- a/ansible/roles/cassandra3/tasks/main.yml
+++ b/ansible/roles/cassandra3/tasks/main.yml
@@ -4,6 +4,17 @@
 
 ########## REDHAT ########## 
 
+- name: check if datastax package is installed
+  package:
+    name: "{{datastax}}"
+    state: present
+  check_mode: true
+  register: cassandra_first_install
+  when: ansible_os_family == "RedHat"
+  tags:
+    - packages
+    - cassandra
+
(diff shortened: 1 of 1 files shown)


djtfmartin:
thanks @vjrj !

vjrj:
You are welcome @djtfmartin !

Issue #308, opened 2019-07-17
nginx_vhost fragments delete fails


vjrj:
Related with #256, updating our spatial service we get this warning:
```
TASK [nginx_vhost : if not using https, delete all https fragments] *************************************************************************************************************************
 [WARNING]: Unable to find '/etc/nginx/vhost_fragments/espacial.gbif.es' in expected paths (use -vvvvv to see paths)
```

After playing a little bit trying to fix the issue I found that we are making the lookup locally:
https://github.com/ansible/ansible/issues/34497#issuecomment-355591155

Some code sample to delete multiple files:
http://www.mydailytutorials.com/ansible-delete-multiple-files-directories-ansible/

ansell:
The ALA do not have any non-HTTPS hosts, so I can imagine we don't use this step ourselves. It may have developed a bug with recent ansible versions or it may have never worked.

vjrj:
It sounds like it never worked... 

Sometimes I test a service without ssl (like in https://github.com/AtlasOfLivingAustralia/spatial-hub/issues/238) or I change to use http/https in new nodes, and I have to delete the fragments directory files manually.

The rest of `with_fileglob` use in `ala-install` seems correct (is used for copy local files).

matthewandrews:
This is completely normal with first-time deploys.  The step is to delete any existing https fragments, which won't be there if it's a first-time deploy.  It's a warning, not an error.  Probably the task name should be edited to make it clearer that this is not a problem.

vjrj:
No worries. Other link to clarify:
https://docs.ansible.com/ansible/latest/plugins/lookup/fileglob.html
"Matching is against local system files on the Ansible controller. To iterate a list of files on a remote node, use the find module."

vjrj:
I think that this can be the cause of some dups re-running ansible:
```
nginx: [emerg] duplicate location "/" in /etc/nginx/sites-enabled/bie.biodiversityatlas.at.conf:253
```

ansell:
It is a known issue with the current nginx vhost fragments construction process that older fragments may be left around if people are re-using hostnames or not setting the `nginx_vhost_fragments_to_clear` array to contain the hosts they need cleared each time.

Adding a new fail task to the `nginx` role that will check whether users have either created the `nginx_vhost_fragments_to_clear` list themselves or they have specified `nginx_vhost_fragments_to_clear_no_warning=True` to override the warning.

PR #404, opened 2020-06-29
issue #308 : Try to support HTTP-only hosts

ansell:
This is an attempt at a bug fix for issue #308 to try to support HTTP-only hosts that may have been HTTPS in the past and which cannot set `nginx_vhost_fragments_to_clear` to remove all old fragments on each ansible run.


ansell:
@vjrj Can you test these changes on an HTTP-only site configuration?

PR #405, opened 2020-06-30
Update trove API url

vjrj:
Update of trove API url after deprecation of previous API. See: https://github.com/AtlasOfLivingAustralia/bie-plugin/issues/196


Rita-C:
Thanks for updating this @vjrj

PR #408, opened 2020-07-28
Added i18n role to basic playbooks

vjrj:
I've just added the `i18n` role to the basic ALA playbooks that support:
https://github.com/living-atlases/ala-i18n

### DIFF
--- a/ansible/ala-demo-2020.yml
+++ b/ansible/ala-demo-2020.yml
@@ -18,6 +18,7 @@
     - nameindex
     - solr7
     - logger-client
+    - i18n
     - biocache-properties
     - { role: biocache-db,      use_cassandra3 : true }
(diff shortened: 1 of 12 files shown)


ansell:
@nickdos @vjrj Before merging this, we need to add the isolation variable to `i18n_package_enabled=False` in each of our internal inventory files that are affected by this so we don't have the `ala-i18n` package installed on our systems until we are ready to use it ourselves.

https://github.com/AtlasOfLivingAustralia/ala-install/blob/master/ansible/roles/i18n/defaults/main.yml#L2

ansell:
Rebased to fix the conflict and merged in https://github.com/AtlasOfLivingAustralia/ala-install/commit/25eb3601abc8035f51a87b69606362fdccb820cc

PR #415, opened 2020-08-14
updated cite-ala url

alexhuang091:
As requested in https://github.com/AtlasOfLivingAustralia/DataQuality/issues/129


alexhuang091:
https://www.ala.org.au/about-the-atlas/terms-of-use/citing-the-atlas/ now redirects to https://support.ala.org.au/support/solutions/articles/6000195866-how-to-cite-the-ala,

so unless it's intentional, we can get rid of the redirect.

PR #421, opened 2020-09-01
Playbooks to support pipelines

djtfmartin:
Includes playbooks for:
* hdfs cluster (but not HA - see https://github.com/AtlasOfLivingAustralia/la-pipelines/issues/132)
* spark cluster
* jenkins with pipeline jobs and jenkins slaves


vjrj:
Following my last comment in my previous review, my main concern is about using docker in the pipelines role (currently only for ala-name-service and ala-sensitive-data-service):
https://github.com/AtlasOfLivingAustralia/ala-install/blob/71545db73d638073be302c813c300bf7b1035482/ansible/roles/pipelines/tasks/main.yml#L104

For instance, nowadays I cannot use docker in my 9 testing VMs I'm using for two remote sessions and my tests.

This is why I created this debian package and suggested they use:
https://github.com/AtlasOfLivingAustralia/ala-namematching-service/pull/12
I can create a similar one for the sensitive-data service.

Also is something that make the maintenance a bit more complicated for small LA portals without too much technical knowledge.

So if we can simplify this role a bit only installing two debs or using it optionally will be more easy for us.

Otherwise, If you feel that it's the moment to start using these docker images in production, I have to mentalize myself to start to document the use and maintenance in others LA portal.

djtfmartin:
@vjrj very happy to have a debian version of SDS akin to the ala-namematcing-service debian package

vjrj (approved):

PR #431, opened 2020-09-28
python 2 dependency compatible with 18.04 and 20.04

vjrj:
I used a wildcard to install some python2*-minimal present in any of the ubuntu releases because python-minimal is not present in 20.04.

### DIFF
--- a/ansible/ala-demo-basic.yml
+++ b/ansible/ala-demo-basic.yml
@@ -2,7 +2,7 @@
   gather_facts: False
   tasks:
   - name: install python 2
-    raw: sudo bash -c "test -e /usr/bin/python || (apt -qqy update && apt install -qy python-minimal)"
+    raw: sudo bash -c "test -e /usr/bin/python || (apt -qqy update && apt-get install -qy 'python2.*-minimal')"
 
 - hosts: all
   roles:
--- a/ansible/ala-demo.yml
+++ b/ansible/ala-demo.yml
@@ -2,7 +2,7 @@
   gather_facts: False
   tasks:
   - name: install python 2
-    raw: sudo bash -c "test -e /usr/bin/python || (apt -qqy update && apt install -qy python-minimal)"
+    raw: sudo bash -c "test -e /usr/bin/python || (apt -qqy update && apt-get install -qy 'python2.*-minimal')"
 
 - hosts: all
   roles:
--- a/ansible/biocache-cli.yml
+++ b/ansible/biocache-cli.yml
@@ -2,7 +2,7 @@
   gather_facts: False
   tasks:
   - name: install python 2
-    raw: sudo bash -c "test -e /usr/bin/python || (apt -qqy update && apt install -qy python-minimal)"
+    raw: sudo bash -c "test -e /usr/bin/python || (apt -qqy update && apt-get install -qy 'python2.*-minimal')"
 
 - hosts: biocache-cli
   roles:
--- a/ansible/biocache-db.yml
+++ b/ansible/biocache-db.yml
@@ -2,7 +2,7 @@
   gather_facts: False
   tasks:
   - name: install python 2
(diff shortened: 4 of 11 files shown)


ansell:
Ansible-2.9 is fully compatible with Python-3. In addition, Python-2 isn't officially supported by Ubuntu-20.04 as it only appears in the `universe` repository which is for unsupported software, so we don't necessarily want it present for security reasons.

It would be ideal if these raw actions were moved from individual playbooks into a single `new-server.yml` or similar playbook so those who want them could choose to install them if necessary.

vjrj:
As there are still LA portals using for instance `xenial` I was trying to modify as less as possible and not to break things that I cannot test.

If you see, python-minimal is in xenial and bionic, but python2*-minimal is in every distro til groovy (20.10).
https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=python%20minimal

This is why I only added a wildcard to the tasks.

As I'm trying to minimize the extras steps necessary to deploy a LA portal, so I'll prefer not to let the user to choose that playbook or that playbook depending on the distribution. In fact I'm testing the same inventories with 18.04 and 20.04.

But returning back to that python task, here is a good summary of options and pros and cons:
https://www.toptechskills.com/ansible-tutorials-courses/how-to-fix-usr-bin-python-not-found-error-tutorial/

In our extra custom playbooks (that I use for setup non ala servers) I use `ansible_python_interpreter=/usr/bin/python3` instead of raw install python-minimal.

I don't like the python3 `ln` option.

Which option do you prefer? To continue with python-minimal or remove it and to start use python3 as interpreter and suppose that python3 is installed?

Now is `auto`
https://github.com/AtlasOfLivingAustralia/ala-install/blob/15e1ba679eaf44977b864343c47ce38b39b52ac4/ansible/roles/common/vars/main.yml#L5

ansell:
We should restrict the actions in the application specific ansible playbooks to those that are mandatory to run the particular application playbook.

The raw command is a helper step that only needs to be run once ever on a particular server. That fits with having it in a once-only run playbook.

An intermediate solution may be to leave the raw command on the demo playbooks, which the ALA doesn't use, and create a new playbook for initially setting up servers. That could include other modifications in the future and would be completely optional for people to run, so we wouldn't need to worry that it had backwards compatibility issues with existing systems, as we do if we keep the raw command in many of the ansible playbooks.

There were initial issues with `ansible_python_interpreter: auto` that required the transition of our ansible roles from using python-2/pip packages to the equivalent python3/pip3 packages, but that was always going to happen with the transition from python-2 to python-3 over time. It has otherwise enabled a smooth transition to using Ubuntu-20.04 for all of our new ALA servers since that change was made.

vjrj:
The `ala-demo*` is used by newcomers deployments to have some sample branding configured.

Many times we have also a single service per VM. See canadensys setup for instance:
2020-07-30-19-31-13-screenshot

Let me summarize: Our current ala-install fails in this initial step in `20.04` using each playbook that uses this initial raw command. It can be solved only adding this wildcard, that I've tested in `18.04` and `20.04` versions in all the modified playbooks (except phylolink and solr6).

If you want I can try to test the these playbooks in new servers without this raw command also in 18.04 and 20.04.

ansell:
My approach may not have been clear. I have opened another pull request with an example of a newserver playbook in #434 . If that pattern looks okay, we would remove all of the raw commands from the ansible playbooks and users could choose to install that playbook first on new servers or leave the default python install if there is one setup.

For Ubuntu-20.04 support I have previously added workarounds to using the Python-3 versions of the custom plugins so it isn't necessarily essential to actually have python-2 on Ubuntu-20.04 for many roles at this point, and we may be able to migrate other roles to Python-3 versions if they haven't been done so already.

ansell:
Closing this per discussion in https://github.com/AtlasOfLivingAustralia/ala-install/pull/434#pullrequestreview-501369752 . Each of the raw python install actions in the application playbooks are being replaced by a single playbook to run once on a new server to install python.

Once the ansiblew method has been updated, these raw actions will be removed.

PR #432, opened 2020-09-28
Generic tomcat using common facts to choose correct one

vjrj:
Testing in 18.04 and 20.04 I use the tomcat generic role with the tomcat version selected in facts.


ansell:
Thanks, this looks good.

PR #434, opened 2020-09-30
Add newserver ansible playbook to run once on new servers

ansell:
Add a newserver ansible playbook to install both python2 and python3 on new servers.

This avoids forcing users to install python2 if they do not want it present on newer systems, and also avoids having to reinstall it in future instances.

Signed-off-by: Peter Ansell <p_ansell@yahoo.com>

### DIFF
--- a/ansible/newserver.yml
+++ b/ansible/newserver.yml
@@ -1,0 +1,14 @@
+---
+# Use this playbook once on a new server to install python2 and python3
+# must pass target var e.g. ansible-playbook newserver.yml -i /path/to/inventories/your-inventory-file -u ${USER} --private-key=~/.ssh/id_rsa --extra-vars 'target=bla-bla'
+- hosts: '{{ target }}'
+  gather_facts: false
+  pre_tasks:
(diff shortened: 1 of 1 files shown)


vjrj (approved):
I trying to install and/or update LA portals with a single command (`./ansiblew --alainstall=../ala-install all`) [[1]](https://github.com/living-atlases/generator-living-atlas#ansible-wrapper) and I don't know the status of the server/s each ansible-playbook will deploy to, if it has python or not installed, etc. So this will make me exec this playbook before each service playbook run. So I don't like it so much.

But I understand that the ALA necessities are different and your uses, infra and times are different. So LGTM.

ansell:
@vjrj Let me know when you have made that addition to the ansiblew method and then we can remove the raw actions from the application playbooks.

vjrj:
@ansell Done. We can remove it. I've also tested with [raw.patch.txt](https://github.com/AtlasOfLivingAustralia/ala-install/files/5340580/raw.patch.txt) (only in `18.04`).

PR #430, opened 2020-09-28
Added gpg missing dependency in nginx role

vjrj:
Both in ubuntu 18.04 and 20.04 I get an error:
```
fatal: [ala-install-test-1]: FAILED! => {"changed": false, "msg": "Failed to find required executable gpg in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin"}
```
during task " Add PPA key for Nginx repository".

I added `gpg` for these base ubuntu images that does not have gpg preinstalled.

Partially mentioned in #372.

PR #435, opened 2020-10-02
Revert "Added gpg missing dependency in nginx role"

ansell:
Reverts AtlasOfLivingAustralia/ala-install#430


ansell:
Reverting a pull request that broke and needs more testing.

Issue #437, opened 2020-10-05
cas5-standalone playbook doesn't install java


vjrj:
Currently `cas5-standalone` playbook doesn't install java so the cas service cannot start in servers with only the CAS service.

I think that this is happening after the the last changes in the `common` role and because of the use of the `include_role`.

It works for me with:
```
diff --git a/ansible/cas5-standalone.yml b/ansible/cas5-standalone.yml
index b37323da..65bfd572 100644
--- a/ansible/cas5-standalone.yml
+++ b/ansible/cas5-standalone.yml
@@ -1,11 +1,8 @@
 - hosts: cas-servers
+  roles:
+    - common
+    - java
   tasks:
-  - name: Include common
-    include_role:
-      name: common
-  - name: Install java
-    include_role:
-      name: java
   - apt_repository:
       repo: ppa:chris-lea/redis-server
       update_cache: yes
```

BTW, I don't understand the necessity of using of `include_role` in cas and spatial playbooks instead of `roles` as the rest of playbooks. I cannot (or I don't know how), for instance, use `--tags properties` from the command line or similar when using `include_role`. Any tip?

cc @sbearcsiro and @ansell .

vjrj:
Solved with #448

PR #448, opened 2020-11-16
CAS playbooks and roles refactor

vjrj:
This PR do a refactorization of CAS playbooks and roles in order to:
- move tasks from playbooks to roles
- use of `roles` instead of `include_role` so we can use `--tags` properly with these playbooks
- maintain the same structure that the rest of playbooks and simplify the cas playbooks

This also solve: https://github.com/AtlasOfLivingAustralia/ala-install/issues/437

I've tested here (ubuntu `18.04` VMs):
https://auth-es.l-a.site/cas/login
https://auth-es.l-a.site/userdetails/
https://auth-es.l-a.site/apikey/

I'm working in this utility:
https://github.com/living-atlases/la-data-generator
that does not generate the properties for the `cas` services because of the current structure of the `cas` playbooks. So this also fix this `cas` properties generation (useful for LA based docker portals).

cc @djtfmartin

### DIFF
--- a/ansible/aws-cas-5.yml
+++ b/ansible/aws-cas-5.yml
@@ -8,195 +8,25 @@
   serial:
   - 1
   - "100%"
-  tasks:
-  - include_role:
-      name: common
-  - apt:
-      name: "{{ packages }}"
-    vars:
-      packages:
-      - mysql-client
-      - python-mysqldb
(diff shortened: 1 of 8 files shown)


djtfmartin (approved):
LGTM

PR #449, opened 2020-11-20
Added when to some handlers so can be skipped with -e skip_handlers=true

vjrj:
Following https://stackoverflow.com/a/57677835 I added some extra var `skip_handlers` to allow prevent some handler to run (for instance if we want to update properties without restart a service).

This should not affect the current usage of `ala-install` and only skip the restart of some services when `-e skip_handlers=true` is added to `ansible-playbook`.


ansell:
@vjrj  The checks in this need to be changed to use `| bool` to reliably convert to booleans in each case. The string based checks only work on some inventory formats (even though this is designed to be used with `--extra-vars` it is best to have it consistent.

PR #461, opened 2021-02-01
apikey-add role to insert apikeys from ansible

vjrj:
As I mentioned in #460 I'm trying to minimize the additional steps required to setup properly a new LA portal: 
https://github.com/AtlasOfLivingAustralia/documentation/wiki/LA-Quick-Start-Guide#post-install
Some of these are the `apikeys` configuration steps:
https://github.com/AtlasOfLivingAustralia/documentation/wiki/API-Keys
I noticed that a pair of new LA portals only completed this post-install step after one year of their initial deployment. 

This new `apikey-add` role can be included in other existing roles and inserts api keys in the `apikey` database if some variables are present and if these apikeys do not exist yet:
```
# Common optional variables:
apikey_def_creator_email=admin@example.com
apikey_def_creator_userid=1

# Variables for some inventory like `bie-hub`:
# Some apikeys generated with `uuidgen` or similar:
bie_index_api_key=bbca0802-4c5a-490c-a0aa-fdd6f02ecbee
speciesList_api_key=ed629e30-2882-48e7-a7d6-704366abdd24
# As we had to insert these keys in the db, these `apikey` db variables should be present for other roles and not only in `cas-servers` group:
apikey_db_name=apikey
apikey_db_username=apikey
apikey_db_password=somepass
# and this if the db location is different from the default ones
apikey_db_hostname=localhost
apikey_db_port=3306
```
This new role is useful for new LA deployments that can automatically configure the ALA apikeys and minimize the effort to deploy a functional LA portal. But also for existing LA portals, in order to have correctly configured these keys, to renew them, and so on.

It's also useful for create apikey apps that are not present in an initial LA deployment. That is, one less confuse additional posinstall step.

As an example I included this new role in `bie-hub` and in `biocache-hub`, but we can use it in the rest of roles/playbooks in the future.


djtfmartin (approved):

vjrj:
Thanks @djtfmartin indeed for the review.

PS: I was thinking in doing a similar thing for IP allowlisting (in logger and userdetails):
https://github.com/AtlasOfLivingAustralia/documentation/wiki/Secure-your-LA-infrastructure#allowlist-ip-address

vjrj:
I didn't notice that I was testing in a testing deployment with CAS in the same VM that bie-hub. In my CI environment with a different setup failed obviously:
```
TASK [apikey-add : ensure target directories exist] ****************************
Wednesday 03 February 2021  12:56:42 +0100 (0:00:00.105)       0:02:12.823 **** 
failed: [ala-install-test-1] (item=/data/apikey/setup) => {"ansible_loop_var": "item", "changed": false, "gid": 0, "group": "root", "item": "/data/apikey/setup", "mode": "0755", "msg": "chown failed: failed to look up user apikey", "owner": "root", "path": "/data/apikey", "size": 4096, "state": "directory", "uid": 0}
```
I'll test better and send a new PR.

PR #472, opened 2021-03-02
mysql common role for 5.7 and 8.0

vjrj:
- Added mysql common role for `5.7` and `8.0` based in common facts and mysql variables depending on the ubuntu version.
- Renamed old mysql role to `mysql-legacy`
- Tested against `u18.04` and `u20.04` servers

cc @nickdos


matthewandrews (approved):
Looks good Vicente.

vjrj:
Thanks @matthewandrews. In some point we can remove mysql 5.7 and 8.0 roles that now are not used.

PR #477, opened 2021-04-14
Replaced location with if $uri to fix #476

### LINKED ISSUE #476: Using CORS with proxy pass in nginx role prevents requested for web fonts from being proxied
To reproduce:
Configure the nginx_vhost role like:
name: add nginx vhost if configured
  include_role:
    name: nginx_vhost
  vars:
    appname: "myapp"
    hostname: "{{ myapp_hostname }}"
    context_path: "{{ myapp_context_path }}"
    tomcat_server_port: "{{ myapp_server_port }}"
    nginx_paths:
      - path: "{{myapp_proxy_target}}//"
        sort_label: "1"
        is_proxy: true
        alias: "default_proxy_all"

Enable CORS by configuring the ginx_cors_origin_regexp variable.

Attempt to serve a web font from the application nginx is proxying:
e.g. GET /assets/my/webfont.woff

Expected:
Status code 200

Actual:
Status code 404 with error in logs indicating the font was attempted to be served from the document root on the file system.

The likely cause is the nested location directive in ala_cors.j2 somehow invalidates the surrounding location / generated for the proxy.

chrisala:


matthewandrews:
This looks good - as discussed it's good to avoid `if` in nginx but there are lots of `if` blocks in that file already, so that's fine here.

Issue #481, opened 2021-05-08
TLS 1.0/1.1 disabled in openjdk 8u292 and newer


vjrj:
@jloomisVCE from Vermont Atlas of Life reported me on Thursday that some of their portal services stopped to work suddenly because a `javax.net.ssl.SSLHandshakeException`.

After some investigation we discovered that `openjdk` was updated by `unattended-upgrades` these days and the problem arose after a restart of their servers. So other portals (like ALA or gbif.es) that autoupdate security packages can be affected these days also. For instance `etckeeper` show this change in `/etc/java-8-openjdk/security/java.security` some days ago in one of our servers:

![2021-08-08-08-05-43-screenshot](https://user-images.githubusercontent.com/180085/117529248-d6690100-afd6-11eb-8118-bd110821a956.png)

Some error include ssl mysql connection errors that can be solved disabling ssl in the mysql connection url with `useSSL=false`, see https://github.com/AtlasOfLivingAustralia/webapi/issues/27#issuecomment-818320684. But other errors (in CAS and geoserver) suggest that this is not sufficient for other services. 

So other option is to re-enable `TLS 1.0/1.1` as described in:
https://aws.amazon.com/blogs/opensource/tls-1-0-1-1-changes-in-openjdk-and-amazon-corretto/

I'm doing some test trying to patch `ala-install` with a workaround and also to detected which services are affected (for now collectory, specieslist, logger, cas, userdetails, apikey, cas-management, alerts and geoserver). 

cc @matthewandrews and @Rita-C

vjrj:
Fixed

PR #482, opened 2021-05-08
Fix for #481 enabling tls 1.0/1.1 for openjdk-8

vjrj:
Fix for #481 following https://aws.amazon.com/blogs/opensource/tls-1-0-1-1-changes-in-openjdk-and-amazon-corretto/ first workaround option.

Maybe we should add this also for openjdk11 but I didn't have the opportunity to test it.


djtfmartin:
thank you @vjrj

vjrj:
Yes @charvolant, sorry, I've just pushed.

PR #520, opened 2021-11-17
Added ansible variables to la-pipelines config

vjrj:
I added some more ansible variables to the local la-pipelines config. I used the values from the:
https://github.com/gbif/pipelines/blob/dev/livingatlas/configs/la-pipelines.yaml
trying to not affect the ALA current inventories.

I added a pair or tags also to continue using the `properties` tag as with other roles.


djtfmartin (approved):

djtfmartin:
thanks @vjrj !

vjrj:
gracias a ti!

PR #523, opened 2021-11-24
Set default value for collectory auth_cookie_name

vjrj:
Similar to other service, just adding some default value as https://github.com/AtlasOfLivingAustralia/ala-install/commit/27d62060935538f77b887ec4e1242ec0402566de fails on new deployments if this var is not defined:

image


djtfmartin:
thanks @vjrj ! - my bad :)

vjrj:
:-)

PR #532, opened 2021-12-16
LA apt repos configurable

vjrj:
This allow to configure stable/unstable repo in apt.gbif.es so we can choose between tagged stable versions of packages and snapshot packages.

So you can select:
``` 
la_apt_repo = 'stable'
``` 

``` 
la_apt_repo = 'unstable'
``` 

Current content of stable:

image

`ala-senstive-data-service` is not there til we have a tagged version with the debian package.

We use a template now because if not several lines are added to the sources without replacing the previous one:
https://github.com/ansible/ansible/issues/58386#issuecomment-505888418

### DIFF
--- a/ansible/roles/la-apt/defaults/main.yml
+++ b/ansible/roles/la-apt/defaults/main.yml
@@ -1,0 +1,1 @@
+la_apt_repo: 'unstable'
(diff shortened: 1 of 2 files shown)


vjrj:
@djtfmartin I think that we can merge and start to use this. The yesterday new stable release of GBIF is builded/published:

image

Issue #539, opened 2022-01-12
bie redirects to bie-index in sites using "/" as context.


vjrj:
It seems that ALA is switching from `bie.ala.org.au/ws` to `bie-ws.ala.org.au` via this commit https://github.com/AtlasOfLivingAustralia/ala-install/commit/719325b7d1170c6427748ebe3d8c14aa21dca4e6 but this breaks sites using `/` as `bie_index_context_path` so in sites with url `bie-ws.example.com` (with `/` context), it species url `bie.example.com` redirect to `bie-ws.example.com`.

This breaks many sites but was discovered by @biodivAtlasAT.

We should do this redirect conditional or optional, but I don't know how yet, maybe with two lists of nginx paths (one conditional).
https://stackoverflow.com/questions/31045106/append-list-variable-to-another-list-in-ansible

cc @matthewandrews

vjrj:
Fixed

PR #540, opened 2022-01-12
Fix for #539

vjrj:
This fix the #539. Test with `/ws` `bie_index_context_path` (ALA use), the bie vhost with the bie-index redirect looks like :

image

Test with empty `bie_index_context_path` (used in several portals of the LA community) and without the redirect:

image

cc @biodivAtlasAT


matthewandrews:
Thanks @vjrj !

vjrj:
Thanks indeed @matthewandrews . We were suffering with this, with unexepected redirects.

Cc @biodivAtlasAT

PR #542, opened 2022-01-17
Conditional artifact for collectory depending on the version

vjrj:
Added some ternary operator in order to start using `collectory` >= `3.0.0` and match the artifact name when not defined.

In this case, we can select the correct artifact only selecting the `collectory` version.

PS: I just added also the `skin_layout` configuration depending on the version


vjrj:
Closing this as I found a bit confusing.

PR #547, opened 2022-02-10
Added extra variables in templates that still are using ALA default values

vjrj:
After #546 I review all our generated configs using `/alaAdmin/viewConfig`pages and I discovered many urls and settings that still are using ALA default values.

This PR adds these variables to the different service config templates. I reviewed to not affect to ALA portal, setting the same defaults values.


djtfmartin:
gracias @vjrj !

PR #558, opened 2022-03-28
Pipelines batch jobs for current slave number  

vjrj:
The PR includes two changes to pipelines jobs for use by the LA community:
- Instead of configure 6 slaves, get the info from jenkins:
image
- Disable docker restart job when is disabled in ansible:
image

Pending: I don't know how to do with this ALA specific job configuration:

image


vjrj:
Should be ok for ALA to do something similar like the [Interpret all job](
https://github.com/AtlasOfLivingAustralia/ala-install/blob/d7b529dd5601afad22e0faee0007154d9e9d31ea/ansible/roles/pipelines_jenkins/templates/jenkins/jobs/Interpret-all/config.xml.j2#L67) to `split` large/small drs? I mean, instead of [this]( https://github.com/AtlasOfLivingAustralia/ala-install/blob/d7b529dd5601afad22e0faee0007154d9e9d31ea/ansible/roles/pipelines_jenkins/templates/jenkins/jobs/Dwc-verbatim-all/config.xml.j2#L76).

djtfmartin:
thanks @vjrj

PR #562, opened 2022-04-18
Update biocollect and ecodata roles for the LA community

vjrj:
Same changes to the `biocollect` and `ecodata` roles, so can be uses by the LA community. This PR includes:

- Allow to configure some ALA variables (like cas, etc)
- Allow to use an alternate port instead of 8080 so can be installed with other services
- some tag addition
- add the `ecodata_api_key` to the `apikey` db if not present.
- avoid the use of `all` as target as this causes problems when we use a unique inventory for all the LA servers/services.


djtfmartin (approved):
LGTM

PR #565, opened 2022-05-02
Use "version" for ala-bie-hub deployment

Rita-C:


vjrj (approved):
Thanks indeed Rita. 

Minor comment: It's not necessary to rename the `bie_hub_artifact` but anyway I configured in our inventories like this for backward compatibility:
 
image

I found some issue that probably you didn't noticed because you used the `ala-bie-hub` previously and I tested in a fresh install. 

The new artifact was installed correctly:

image

by I detected some issues, like the branding configuration is ignored:

image

and also the auth service configuration sends me to:

https://auth.ala.org.au/cas/login?service=http%3A%2F%2Fdevt.ala.org.au%3A8080%2Fadmin

I think that the cause is that the configuration is copied to `bie_hub`variable (in our case `/data/ala-bie`), see:
https://github.com/AtlasOfLivingAustralia/ala-install/blob/c0a71a6f65385396cae133319fedd19fdcb71fe6/ansible/roles/bie-hub/tasks/main.yml#L18

But the new application, expect it in `/data/ala-bie-hub`:

https://github.com/AtlasOfLivingAustralia/ala-bie-hub/blob/c658f4758fd630b5e80cd6b0c988d14016d93803/grails-app/conf/application.yml#L41

so gets the default grails values.

So maybe we have to configure in our inventories `bie-hub` correctly depending on the version.

Rita-C:
> Thanks indeed Rita.
> 
> Minor comment: It's not necessary to rename the `bie_hub_artifact` but anyway I configured in our inventories like this for backward compatibility:
> 
> image
> 
Thanks, I changed it back to `bie_hub_artifact`

Rita-C:
Thanks very much for your review and QA @vjrj , appreciated!

Issue #602, opened 2022-07-26
The geoserver extensions install many times fails with timeout


vjrj:
> TASK [install geoserver extensions] ******************************************************************************************************************************************************
changed: [espacial.gbif.es] => (item=https://download.sourceforge.net/project/geoserver/GeoServer/2.14.1/extensions/geoserver-2.14.1-pyramid-plugin.zip)
failed: [espacial.gbif.es] (item=https://download.sourceforge.net/project/geoserver/GeoServer/2.14.1/extensions/geoserver-2.14.1-vectortiles-plugin.zip) => {"ansible_loop_var": "item", "changed": false, "item": "https://download.sourceforge.net/project/geoserver/GeoServer/2.14.1/extensions/geoserver-2.14.1-vectortiles-plugin.zip", "msg": "Failure downloading https://download.sourceforge.net/project/geoserver/GeoServer/2.14.1/extensions/geoserver-2.14.1-vectortiles-plugin.zip, Request failed: <urlopen error _ssl.c:1114: The handshake operation timed out>"}

As [unarchive in this task](https://github.com/AtlasOfLivingAustralia/ala-install/blob/188f97fb4acc45243ec79b9259050247e329f255/ansible/roles/geoserver/tasks/geoserver.yml#L136) does not have a timeout param to configure, it's better to use `get_url` with a greater timeout and `unarchive` later.

Issue #598, opened 2022-07-14
Ubuntu 20.04 current issues


vjrj:
- [x] spatial, images, doi: postgresql/postgis package missing
- [x] cas:  `Unknown system variable 'tx_isolation'`  with `5.3.12-2` and `6.5.6-1` (https://github.com/AtlasOfLivingAustralia/ala-cas-5/issues/50)
- [x] `ERROR 1267 (HY000) at line 2 in file: '/data/userdetails/setup/auth-ip.sql': Illegal mix of collations (utf8mb4_unicode_ci,IMPLICIT) and (utf8mb4_0900_ai_ci,IMPLICIT)` in `userdetails`.
- [x] spatial: `The target XML source '/data/geoserver_data_dir/gwc-gs.xml' does not exist.`
- [x] Mysql 8 login issues #604 
- [ ] postgresql not available for ubuntu focal (#920)

vjrj:
The last one is caused by https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=956804 so `GEOSERVER_DATA_DIR` is not setted correctly.

vjrj:
Unchecking pg part, as #600 was reverted

PR #603, opened 2022-07-26
Use of JAVA_OPTS instead of CATALINA_OPTS. Fix for #602

vjrj:
Following https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=956804 fix for task of #598 and #602.

### DIFF
--- a/ansible/roles/geoserver/tasks/ecodata-geoserver.yml
+++ b/ansible/roles/geoserver/tasks/ecodata-geoserver.yml
@@ -16,7 +16,7 @@
 - name: geoserver for ecodata - configure geoserver data dir
   lineinfile:
     dest={{tomcat_conf}}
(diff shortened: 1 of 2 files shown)


sbearcsiro:
The Debian maintainer seems to not understand `CATALINA_OPTS`.  `CATALINA_OPTS` is correct AFAIK.

`CATALINA_OPTS` is used only for `catalina.sh` `run`, `start` and `debug`, so you're not setting, eg `Xmx=4G` and `Xms=4G` when running `catalina.sh stop` (which launches a separate jvm process with `JAVA_OPTS` that stops tomcat.  Or at least it did back in the day).

The above still holds for Tomcat 9, I've just checked the package locally.

sbearcsiro:
Seems like we could uselineinfile: {

vjrj:
Thanks @sbearcsiro for the clarifications. Anyway, if you check `ala-install` we only use `CATALINA_OPTS` to set the `GEOSERVER_DATA_DIR` and `GEOWEBCACHE_CACHE_DIR`. So I think that don't merit the effort (to send patch to debian) if we can accomplish the same using `JAVA_OPTS` in the geoserver install.

sbearcsiro:
Thanks @vjrj, you're right, it's not setting memory limits or anything so this change is fine IMHO.

PR #606, opened 2022-08-12
CAS backwards compatibility with pre OIDC versions

vjrj:
This PR allows to still install previous versions of CAS with ala-install. For that:

1. we require the install of java 11 (or 8) depending on each component version
1. we use previous ala-install (2.1.9) configurations of each cas components for that versions and new configurations for recent versions
1. we use jar of wars also depending of the version 

I'm not sure of this PR. I started to do point 1) and continued with the rest. Feel free to reject it.


djtfmartin:
gracias @vjrj

PR #613, opened 2022-09-02
Revert "postgresql versions moved to common facts"

sbearcsiro:
Reverts AtlasOfLivingAustralia/ala-install#600

This change forced installation of different postgres versions than were already installed and is not backwards compatible, needs to be reworked so postgres version isn't defined via set facts which has higher priority than basically everything else.


vjrj:
Maybe addtional variables in `set_facts` https://github.com/AtlasOfLivingAustralia/ala-install/blob/35c1baa05a84cd7e358242899ada132cbe8a8901/ansible/roles/common/tasks/setfacts.yml#L137
can help in this case. What do you thing @sbearcsiro?

PR #624, opened 2022-09-23
Allow to config logging in spark and la-pipelines

vjrj:
This PR allows to customize the la-pipelines `logback.xml` config and the spark `log4j` config. 

I used the `log4j.properties.template` provided by spark but using `WARN` as `log4j.rootCategory` as INFO was too verbose in recent versions of la-pipelines. See https://github.com/GBIFes/portal-datos-gbifes/issues/41


djtfmartin:
thanks @vjrj !

PR #632, opened 2022-11-22
copying apiKey to bioache-service requets headers when jwt_enabled i…

sughics:
…s true for backwards compatibility with legacy usages of apiKey in url params


sughics:
Closing PR pending further decisions.

PR #635, opened 2022-11-28
Events prod deployment

djtfmartin:
Work for https://github.com/AtlasOfLivingAustralia/extended-data-model/issues/80


ansible/roles/events/files/es2vt.env
```diff
@@ -0,0 +1,8 @@
+event:
+  hosts: [http://host.docker.internal:9200]
+  index: event
+  requestTimeout: 60000
+  maxRetries: 3
+  maxResultWindow: 100000
+port: 4002
+apiKey: my-api-key
```
vjrj, line 8: Same here, better to /templates/ and with a api_variable

ansible/roles/events/tasks/main.yaml
```diff
@@ -0,0 +1,145 @@
+- include: ../../common/tasks/setfacts.yml
+  tags:
+    - docker
+    - update-images
+    - events
+    - service
+    - docker-config
+    - events_vhost
+
+- name: Add Docker GPG apt Key
```
vjrj, line 10: Just a comment. I think that this docker part merits to be moved to a reusable new `docker` role (as tomcat/nginx/mysql/etc). This way we can reuse it.

ansible/events.yml
```diff
@@ -0,0 +1,26 @@
+- hosts: events
+  roles:
+    - common
+    - webserver
+    - events
+  vars:
+    docker_dir: "/data/docker"
+
+- hosts: elasticsearch
```
vjrj, line 9: As other services uses elasticsearch too, maybe this group should be renamed to events-elasticsearch to avoid future conflicts

ansible/roles/events/templates/graphql-api.env
```diff
@@ -0,0 +1,15 @@
...
+apiEsKey: {{ es_api_key }}
+es2vt: http:///es2vt:{{ es2vt_port | default('4002') }}
+apiDownloads: "{{ downloads_url }}/dataset/{datasetKey}"
+datasetEml: "{{ collectory_url }}/ws/eml/{datasetKey}"
+alaNamematching: "{{ namesearch_url }}/api"
+ala:
+  namematching: "{{ namesearch_url }}/api"
+  bie: {{ bie_service_url }}
+  biocache: {{ biocache_ws_url }}
+  images: {{ media_store_url }}
```
vjrj, line 15: `media_store_url` is used in other services without the /ws/ so will be great to user a different name to avoid conflicts.

ansible/events.yml
vjrj: better `events_elasticsearch` following https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#creating-valid-variable-names

vjrj (approved):
More than the minimal comments I did, LGTM. Feel free to solve it now or later.

PR #637, opened 2022-12-14
OIDC extra variables for bie. More configurable urls

vjrj:
This PR externalize some extra url variables and then they can be configured by other LA portals (see ala.org.au variables):

image
image

Also allow to fine tune OIDC and related variables.


vjrj:
I merge this as allow setting these variable to non ALA portals and .

PR #638, opened 2022-12-19
automate ecodata daily/weekly/monthly backup management

joe-lipson:
Untested. just checking if I got the right idea

does `mongodump` create the dump directory? Wondering if that should be cleaned up after the backup is all tgziped up

### DIFF
--- a/ansible/roles/ecodata/templates/backup-ecodata.sh
+++ b/ansible/roles/ecodata/templates/backup-ecodata.sh
@@ -1,8 +1,33 @@
 #!/bin/bash
-mkdir -p {{data_dir}}/backups
+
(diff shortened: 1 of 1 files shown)


chrisala (approved):
Looks good to me Joe, although I would like to keep more backups if possible, minimum of two weeks of daily backups as it's common to get bug reports more than a week old and being able to restore the state of the system then is useful.
Similarly, I've been asked to restore files > 1 year old before due to compliance audits so would like to keep a couple of years of monthly backups at a minimum

joe-lipson:
Yep sure we can make those retention values whatever is required. I'll change it to 2 weeks on dailys and a year of monthlys but feel free to change them as required

PR #640, opened 2022-12-22
data-quality-filter-service oidc and other variables

vjrj:
- Added new oidc related variables to `dq-filter-service` role.
- Added extra named secret/clientId variables for LA portals with single inventories.
- Allow to enable/disable oidc for LA portals with not yet migrated auth systems
- Some extra API configuration variables


vjrj:
I merge this as allow setting these variable to non ALA portals and it defaults to current ALA values.

PR #642, opened 2023-01-10
update biocache download purge older than 90 days --> 30 days

joe-lipson:


vjrj:
I think we can merge this, in order to not have too much PRs open.

PR #643, opened 2023-01-13
install jdk11 on the digivol server

joe-lipson:
Had to add a `- java` tag to the set_fact in the Java role to get it working. The install command immediately after that uses that variable and is also tagged a java


vjrj:
We can merge this, in order to not have too much PRs open.

matthewandrews:
partly duplicates changes in [another PR](https://github.com/AtlasOfLivingAustralia/ala-install/pull/679) but that's ok

Issue #641, opened 2022-12-30
Start of tomcat9 in ubuntu 18.04 fails because of JAVA_OPTS


vjrj:
Installing recent collectory and biocache-store in ubuntu 18.04 require tomcat9 but fails to start because of the line:

` JAVA_OPTS="${JAVA_OPTS} -Xmx4g -Xms4g -Xss256k -Djava.awt.headless=true -Dlog4j2.formatMsgNoLookups=true"`

In `/etc/default/tomcat9`.

service tomcat9 status error:
 
```
   Active: failed (Result: exit-code) since Fri 2022-12-30 08:43:45 UTC; 1min 30s ago
     Docs: https://tomcat.apache.org/tomcat-9.0-doc/index.html
  Process: 11601 ExecStart=/bin/sh /usr/libexec/tomcat9/tomcat-start.sh (code=exited, status=1/FAILURE)
Dec 30 08:43:45 ala-install-test-1 tomcat9[11601]: NOTE: Picked up JDK_JAVA_OPTIONS:  --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.rmi/sun.rmi.transport=ALL-UNNAMED
Dec 30 08:43:45 ala-install-test-1 tomcat9[11601]: Error: Could not find or load main class ${JAVA_OPTS}
Dec 30 08:43:45 ala-install-test-1 tomcat9[11601]: Caused by: java.lang.ClassNotFoundException: ${JAVA_OPTS}

```

Workaround while ala-install is patched: to use something like `JAVA_OPTS_INI`  in first place and include it later:
```
JAVA_OPTS_INI="-Djava.awt.headless=true -XX:+UseG1GC"
JAVA_OPTS="${JAVA_OPTS_INI} -Xmx4g -Xms4g -Xss256k -Djava.awt.headless=true -Dlog4j2.formatMsgNoLookups=true"
```
instead of:
```
JAVA_OPTS="-Djava.awt.headless=true -XX:+UseG1GC"
JAVA_OPTS="${JAVA_OPTS} -Xmx4g -Xms4g -Xss256k -Djava.awt.headless=true -Dlog4j2.formatMsgNoLookups=true"
```

Or use ubuntu 20.04.

PR #644, opened 2023-02-14
Fix for #641 and added debug

vjrj:
This is a fix of #641, adding the default options of debian instead of the variable `$JAVA_OPTS` only for ubuntu `18.04` when using `tomcat9`.


vjrj:
Merged as I'm sick of doing this manually :-)

PR #650, opened 2023-03-16
Allow to enable/disable oidc in image service

vjrj:
I added a flag just similar to:

https://github.com/AtlasOfLivingAustralia/ala-install/pull/640/files
https://github.com/AtlasOfLivingAustralia/ala-install/pull/639/files


vjrj:
I merged it as is quite safe.

PR #658, opened 2023-04-25
Allow to show vernacular names optionally in biocache

vjrj:
This PR just adds the new variable introduced by https://github.com/AtlasOfLivingAustralia/biocache-hubs/pull/562 so it can be configurable.


vjrj:
Merged as the default value match the current ALA use.

PR #666, opened 2023-06-19
update version of pg client

joe-lipson:
The pg client needed to be updated to v15 to be able to connect to the new RDS instance.

Unsure if this is correct, does this install the server as well?


vjrj:
Yes, this install the server too.

matthewandrews:
I think this PR can be closed - volunteer/digivol is all remote db now, covered in a different PR.

PR #691, opened 2023-10-05
change include to include_tasks

matthewandrews:
`include` is deprecated from ansible 2.4. It appears to behave like the new `import_*` (preprocessing) as opposed to the new `include_*` (runtime parsing).

### DIFF
--- a/ansible/roles/bie-hub/tasks/main.yml
+++ b/ansible/roles/bie-hub/tasks/main.yml
@@ -198,9 +198,9 @@
     - bie
 
 - name: Include docker tasks
-  include: docker-tasks.yml
+  include_tasks: docker-tasks.yml
   when: deployment_type == 'swarm'
 
 - name: Include VM tasks
-  include: vm-tasks.yml
+  include_tasks: vm-tasks.yml
   when: deployment_type == 'vm'
--- a/ansible/roles/bie-index/tasks/main.yml
+++ b/ansible/roles/bie-index/tasks/main.yml
@@ -173,9 +173,9 @@
     - properties
 
 - name: Include docker tasks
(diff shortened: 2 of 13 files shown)


vjrj (approved):
Yes, include is deprecated in recent versions. I didn't realized I was using it here.

djtfmartin (commented):
LGTM

vjrj:
As a side effect of this, now we cannot select tags inside the `include_tasks`. 

An option is to use `import_tasks` instead (the number or skipped tasks will be increased). Other option is to add tags to the `include_tasks` (less useful because the internal tags will be ignored), see [the tags that I just added here](https://github.com/AtlasOfLivingAustralia/ala-install/commit/e11d8a8f886671627253165d89bdb3798b82106a).

matthewandrews:
Ah, that's not good. The original reason we had a problem with `include` (or `import_tasks`) is that it required people to have ansible `docker_stack_module` installed, even if only using VMs.

vjrj:
Strange, because I didn't installed the `docker_stack_module`.

matthewandrews:
Oh - possibly another module was required.  The error we were getting was like this:
```
ERROR! couldn't resolve module/action 'docker_stack'. This often indicates a misspelling, missing collection, or incorrect module path.

The error appears to be in 'ala-install/ansible/roles/docker-common/tasks/stack-create.yml': line 12, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: Remove Docker Swarm stack
  ^ here
```

vjrj:
After searching my shell history, but I don't find nothing related with some module installation.

I'll test it in a clean environment.

PR #695, opened 2023-10-05
Create bie and bie-offline collection if not exists

vjrj:
As commented with @djtfmartin, in `solrcloud-config` role `bie` and `bie-offline` collections are not created, only `biocache`.

But in `solr` old roles all cores are created.

This PR add both `bie` collections if they don't exist (similarly to `biocache`).

I added some tags, and remove some `where` not necessary in docker-swarm deployment.

### DIFF
--- a/ansible/roles/solrcloud/tasks/docker-facts.yml
+++ b/ansible/roles/solrcloud/tasks/docker-facts.yml
@@ -18,6 +18,7 @@
     - docker
     - docker-service
     - docker-compose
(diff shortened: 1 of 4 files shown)


djtfmartin (approved):
LGTM

matthewandrews:
I'm happy to merge this now - we might look at adding more conditionality later, as ALA's own setup currently has biocache and BIE on separate solr clusters.

PR #696, opened 2023-10-05
Playbooks to deploy by-type

vjrj:
I added some new playbooks to deploy on VMs or other deployments types based a `deployment_type` variable.

I'm not 100% happy with the solution as I prefer the something like "include" but for the moment it works. Maybe it can be improved in future ansible releases:
https://github.com/ansible/ansible/issues/34281

As this only add new playbooks it does not affect a previous uses.

### DIFF
--- a/ansible/auth-by-type.yml
+++ b/ansible/auth-by-type.yml
@@ -1,0 +1,11 @@
+---
+- hosts: cas-servers
+  gather_facts: false
+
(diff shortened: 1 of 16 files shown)


djtfmartin (approved):
LGTM

matthewandrews:
Looks good, merging.

PR #698, opened 2023-10-06
Trying to fix 694

### LINKED ISSUE #694: Undefined variable 'groupId' in tomcat_deploy
We introduced a new issue with #690 , I think (or maybe we should update our inventories):

```
TASK [biocache3-service : download from maven repo https://nexus.ala.org.au/service/local/artifact/maven/redirect?r=public&g=au.org.ala&a=biocache-service&v=3.1.5&e=war&c=] ***
Thursday 05 October 2023  12:27:04 +0200 (0:00:00.505)       0:07:52.449 ****** 
fatal: [ala-install-test-1]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'groupId' is undefined\n\nThe error appears to be in '/data-additional/jenkins/ala-install-test/ala-install/ansible/roles/tomcat_deploy/tasks/main.yml': line 61, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: download from maven repo {{ war_url }}\n  ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes. Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}
```

cc @matthewandrews

vjrj:
This PR tries to fix #694 with a fallback.

### DIFF
--- a/ansible/roles/geoserver/defaults/main.yml
+++ b/ansible/roles/geoserver/defaults/main.yml
@@ -1,8 +1,8 @@
-#groupId: "org.geoserver.web"
-#version: "{{ geoserver_version | default('2.14.1') }}"
-#artifactId: "gs-web-app"
-#packaging: "war"
-#classifier: "war"
+groupId: "org.geoserver.web"
+version: "{{ geoserver_version | default('2.19.2') }}"
+artifactId: "gs-web-app"
+packaging: "war"
+classifier: "war"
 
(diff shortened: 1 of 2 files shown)


matthewandrews (approved):
Looks good, thanks

PR #702, opened 2023-10-09
lists 4.4.0 database update

qifeng-bai:
SQL scripts to create / update tables

### DIFF
--- a/ansible/roles/species-list/files/v4.4.0_update.sql
+++ b/ansible/roles/species-list/files/v4.4.0_update.sql
@@ -1,0 +1,22 @@
+#Create a matched_species table
+    CREATE TABLE `matched_species` (
+           `id` int NOT NULL AUTO_INCREMENT,
+           `taxon_concept_id` varchar(255) DEFAULT NULL,
+           `scientific_name` varchar(255) NOT NULL,
+           `scientific_name_authorship` varchar(255) DEFAULT NULL,
+           `vernacular_name` varchar(255) DEFAULT NULL,
+           `kingdom` varchar(255) DEFAULT NULL,
+           `phylum` varchar(255) DEFAULT NULL,
+           `taxon_class` varchar(255) DEFAULT NULL,
+           `taxon_order` varchar(255) DEFAULT NULL,
+           `family` varchar(255) DEFAULT NULL,
+           `genus` varchar(255) DEFAULT NULL,
+           `taxon_rank` varchar(255) DEFAULT NULL,
+           `version` BIGINT NOT NULL DEFAULT 0,
+           PRIMARY KEY (`id`,`scientific_name`),
+           UNIQUE KEY `id_UNIQUE` (`id`)
+    )
+
+# Link the new created matchedSpecies table to speciesListItem table:
+    ALTER TABLE `specieslist`.`species_list_item`
+        ADD COLUMN `matched_species_id` INT DEFAULT NULL


ansible/roles/species-list/files/v4.4.0_update.sql
```diff
@@ -0,0 +1,22 @@
...
+           `vernacular_name` varchar(255) DEFAULT NULL,
+           `kingdom` varchar(255) DEFAULT NULL,
+           `phylum` varchar(255) DEFAULT NULL,
+           `taxon_class` varchar(255) DEFAULT NULL,
+           `taxon_order` varchar(255) DEFAULT NULL,
+           `family` varchar(255) DEFAULT NULL,
+           `genus` varchar(255) DEFAULT NULL,
+           `taxon_rank` varchar(255) DEFAULT NULL,
+           `version` BIGINT NOT NULL DEFAULT 0,
+           PRIMARY KEY (`id`,`scientific_name`),
```
joe-lipson, line 16: Does the primary key need to be composite? it seems redundant as id is auto incrementing

joe-lipson:
looks good, assuming this will be run manually. for an easy rollback it would be good to create a ansible/roles/species-list/files/v4.4.0_rollback.sql file to quickly undo the changes if needed.

qifeng-bai:
> looks good, assuming this will be run manually. for an easy rollback it would be good to create a ansible/roles/species-list/files/v4.4.0_rollback.sql file to quickly undo the changes if needed.

We used to backup the whole database when we need to update DB.  Restore DB from backup if any failures

ansible/roles/species-list/files/v4.4.0_update.sql
qifeng-bai: It is for indexing the scientific_name.  I quickly scan the docs of Mysql.  It says creating an index via adding it to primary key.  Correct me if I am wrong

joe-lipson: The way composite keys work a query for just the scientific name will not use the index, it will have to do a full table scan. Scientific name can be added as a 2nd index

vjrj:
I arrive late to the party, but I think that this kind of sql files without any task associated suggest me that we should exec this sql in some point manually, and this is not a good devops practice. Also this is a bit more complicate to do in docker environments.

For instance, the `ONLY_FULL_GROUP_BY.sql` file in the same directory has a task to exec it.

Anyway, we should move these db related tasks to something like `liquibase` or `flyway` directly in the code. These tools also provide rollback functionality if something does not work as expected. And also these db changes are well in sync with the code itself (new fields or indexes are added when they are really needed, for give an example).

I can help with that.

cc @djtfmartin

djtfmartin:
I support the idea of moving this into liquibase ( as we have done in collectory and other components) as this helps with containerisation as well, but i too am late to the party : )

@vjrj  - would you be able to do a PR moving to liquibase ?

@joe-lipson @qifeng-bai are you happy with this ?

vjrj:
Yes, I can do it if we agree.

qifeng-bai:
Yes, I agree

joe-lipson:
Agree in principal that any manual tasks are not good DevOps practice, but we are a long way away from any good DevOps practice in the way we install or manage our systems. It's a script thats checked in which is at least a good start, even if it's initiated manually it's an automated script thats checked in and reviewed, there should be a rollback script as well. It will only be run once, I dont see the value in introducing a migration tool into this workflow at this stage. Flyway or liquidbase can be part of a new system.

This change needs to go out, I think we should run this script manually for now, maybe we can move it to a different location, arguably it doesnt belong in the Ansible files directory

vjrj:
One thing it's the decision to add liquibase or flyway to the ALA modules that still do not use them (lists, doi, alerts, for instance) and other thing is this sql that is clear that is needed to run in `4.4.0`.

Sometimes these changes are not documented or added to a sql file, and they are done in production directly causing sooner or later headaches.

In my opinion, the idea of using a migration tool to track a database schema is as fundamental as having a version control system for our code. Again, I can help when you decide to add it.

One extra recommendation is what we did in collectory: to compare the production schema with the development schema (liquibase has a [diff tool](https://docs.liquibase.com/commands/inspection/diff.html)), to detect these undocumented db changes. Something that [we didn't in spatial](https://github.com/AtlasOfLivingAustralia/spatial-service/issues/213) and probably is the cause of the boot issue mentioned and the corresponding headache🤦🏻‍♂️ .

djtfmartin:
thanks @joe-lipson  - agree the change needs to go out and we can run manually now.
I should have made it clear I was thinking this refactoring would happen after that.

joe-lipson:
Thanks @vjrj and @djtfmartin In agreement about using a migrations tool and it's importance, just needed to get this out. That actually raises something I've been thinking about more generally regarding our engineering principals, I've come across the same problems with differences in test / prod + many other places where we can improve. I think we're working towards the same goals but it would be useful to put together a list of platform features that we are working towards. Not tech specifics but general things like, everything is IaC, automation - no manual configs / changes, all services can autoscale, auto heal, HA/FT etc. Anyway this PR isn't the place for it but I'll make a start on a general list

PR #734, opened 2023-11-21
Allow better sds configuration in sensitive-data-service

vjrj:
This PR allows to configure better the `sds-config.properties`  of the `sensitive-data-service`, if not, it uses the default ALA values.

Before:

image

After:

image

Thanks to @shahmanash to point to the issue and the fix.

**Update:**

I also moved sds-config to common tasks to configure it also in vm based deployments.


vjrj:
This PR adds some variables with the ALA default values, to allow their configuration and refactor some tasks. I think is safe to merge too.

cc @matthewandrews

matthewandrews:
Thanks @vjrj

PR #756, opened 2023-12-15
have tomcat restart on failure

joe-lipson:
Related to https://github.com/AtlasOfLivingAustralia/ala-install/pull/755

For services that use tomcat to run, have them restart on failure also. 

eg. alerts, lists, spatial, collections


matthewandrews (approved):

matthewandrews (commented):
I think Restart value should be on-failure (hyphen)

joe-lipson:
> I think Restart value should be on-failure (hyphen)

thanks, thats fixed!

Issue #794, opened 2024-03-28
Error in pg_instance role enabling extensions 


vjrj:
Installing `spatial` I get this error:

```
TASK [pg_instance : enable extensions "['postgis', 'postgis_topology', 'uuid-ossp']" for "layersdb" - local] ************************************************************************************************************************************************************************************************************************************************
task path: /home/vjrj/proyectos/gbif/dev/ala-install/ansible/roles/pg_instance/tasks/main.yml:98

The full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_postgresql_ext_payload_unb9p1xp/ansible_postgresql_ext_payload.zip/ansible_collections/community/general/plugins/modules/postgresql_ext.py", line 344, in main
  File "/tmp/ansible_postgresql_ext_payload_unb9p1xp/ansible_postgresql_ext_payload.zip/ansible_collections/community/general/plugins/modules/postgresql_ext.py", line 289, in ext_get_versions
  File "/tmp/ansible_postgresql_ext_payload_unb9p1xp/ansible_postgresql_ext_payload.zip/ansible_collections/community/general/plugins/modules/postgresql_ext.py", line 292, in <listcomp>
  File "/usr/lib/python3.8/distutils/version.py", line 64, in __gt__
    c = self._cmp(other)
  File "/usr/lib/python3.8/distutils/version.py", line 337, in _cmp
    if self.version < other.version:
TypeError: '<' not supported between instances of 'str' and 'int'
failed: [gbif-es-espacial-2024-1] (item=postgis) => {
    "ansible_loop_var": "item",
    "changed": false,
    "invocation": {
        "module_args": {
            "ca_cert": null,
            "cascade": false,
            "db": "layersdb",
            "ext": "postgis",
            "login_host": "",
            "login_password": "",
            "login_unix_socket": "",
            "login_user": "postgres",
            "name": "postgis",
            "port": 5432,
            "schema": null,
            "session_role": null,
            "ssl_mode": "prefer",
            "state": "present",
            "trust_input": true,
            "version": null
        }
    },
    "item": "postgis",
    "msg": "Database query failed: '<' not supported between instances of 'str' and 'int'"
}
redirecting (type: modules) ansible.builtin.postgresql_ext to community.general.postgresql_ext
```

Similar to [this one](https://github.com/ansible-collections/community.general/pull/1104) that is solved in recent versions.

adam-collins:
[...]

PR #795, opened 2024-03-28
Use shell instead of postgresql_ext to enable extensions

vjrj:
Workaround for #794.


adam-collins:
Fixes the error for me.

PR #803, opened 2024-05-02
Unpin docker versions

vjrj:
Unpin docker versions to allow to use the same version variables as the VM tasks and use tags of 
https://hub.docker.com/u/atlasoflivingaustralia

For instance:
https://hub.docker.com/r/atlasoflivingaustralia/collectory/tags


vjrj:
Merged this as is does not affect to ALA and permits me to propose new changes.

PR #807, opened 2024-06-07
Revert relative symlink to sites-available for docker environments

vjrj:
This allows to use  nginx confs generated by ansible outside the containers but it works too in traditional VM environments.

This was introduced by 9b58a27d6f2650d78ebdf83f28e763046167427a and lost in aacd03a1586a328f8291e20f77f796ab20b7e4ad  for some reason.

### DIFF
--- a/ansible/roles/nginx_vhost/tasks/main.yml
+++ b/ansible/roles/nginx_vhost/tasks/main.yml
@@ -526,7 +526,7 @@
 - name: symlink vhost to sites-enabled
   file:
     state: link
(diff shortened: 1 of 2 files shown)


vjrj:
Merging this, as was causing links and references in nginx docker based deployments to looks like:
```
        include /data/nginx/conf.d/ala_cors_cas-management;
```
as the nginx (and other configurations) are generated in `/data` and mounted as volumes. With this PR:
```
        include ../conf.d/ala_cors_cas-management;
```

PR #813, opened 2024-06-24
Updates for image service 3

sbearcsiro:
- Add support for datadog jmx to exec jar roles
- Fix datadog sql role for non local database
- Remove using the CMS collector by default and replace with jvm extra params for clients if required

### DIFF
--- a/ansible/image-service.yml
+++ b/ansible/image-service.yml
@@ -17,7 +17,6 @@
     - image-service
     - { role: datadog_sql_user, become: yes, become_user: postgres, when: datadog_monitoring is defined }
     - { role: datadog.datadog, when: datadog_monitoring is defined }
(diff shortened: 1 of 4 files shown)


yasima-csiro (approved):

joe-lipson:
looks good, I think thats better to inject the tracer and params into the JVM options in the exec-jar role, the datadog_apm role is a bit of a hack. Could there be some other services that are still using the datadog_apm role and also using exec-jar? That could cause problems. I'll look into what else is using it, the old role should be removed, also the vars that were used in it can be cleaned up https://github.com/AtlasOfLivingAustralia/ansible-inventories/blob/master/image-service/image-service-test-cognito.yml#L171-L176

joe-lipson (approved):

PR #837, opened 2024-09-09
Update newserver.yml indentation

vjrj:
There is an indentation issue in `newserver.yml` that gives this error:
```
ERROR! We were unable to read either as JSON nor YAML, these are the errors we got from each:
JSON: Expecting value: line 1 column 1 (char 0)

Syntax Error while loading YAML.
  mapping values are not allowed in this context

The error appears to be in '/data-additional/jenkins/ala-install-test/ala-install/ansible/newserver.yml': line 9, column 10, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  - name: Get the distribution version
      raw: lsb_release -r | awk '{print $2}'
         ^ here
```

### DIFF
--- a/ansible/newserver.yml
+++ b/ansible/newserver.yml
@@ -6,9 +6,9 @@
   gather_facts: false
   pre_tasks:
   - name: Get the distribution version
-      raw: lsb_release -r | awk '{print $2}'
-      register: os_version
-      changed_when: false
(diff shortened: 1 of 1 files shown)


matthewandrews (approved):

matthewandrews:
Oh sorry - merging that now

PR #848, opened 2024-09-19
Add java 17

yasima-csiro:


matthewandrews (approved):

sbearcsiro (commented):
I think this is almost ok for right now - we should probably introduce a default variable of use_openjdk17 = false so you don't need to define it in every single inventory.

However, with a view on the future, I think we should move away from openjdk flags and replace it with a java_version= ie an enum or ansible equivalent of of 8, 11, 17, 21, 25, etc.  We should also probably enumerate the jvm distribution as well, ie java_distro= openjdk, azul, adoptium, corretto (🤮), etc, even though openjdk is the only one supported right now.

vjrj:
I agree with @sbearcsiro . I get this error just testing (without modifying inventories):
```
TASK [java : Show java version] ************************************************
fatal: [ala-install-test-1]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'use_openjdk17' is undefined. 'use_openjdk17' is undefined\n\nThe error appears to be in '/data-additional/jenkins/ala-install-test/ala-install/ansible/roles/java/tasks/main.yml': line 4, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"Show java version\"\n  ^ here\n"}
```

yasima-csiro:
Thanks both. I will set up a default variable of use_openjdk17 = false. 

Regarding the Simon's other comment, I will create a ticket for that and work on separately.

PR #849, opened 2024-09-25
Add default HTTP user agents

sbearcsiro:
Adds a default HTTP User Agent via the Java http.agent system property


sbearcsiro:
@joe-lipson here's something I was toying around with for user agents on the exec jar role - it should work for HTTPUrlConnection and Apache HTTP Client (as long as the client is constructed with defaults) based requests but not, afaik, okhttp based ones.  Should be easy enough to add this to tomcat and the other non spring boot exec-jar roles

joe-lipson (approved):

PR #859, opened 2024-12-02
specieslist-webapp (test) - fix for incorrect webservices creds

nickdos:
See Slack thread for details  - https://atlaslivingaustralia.slack.com/archives/G0104JNLYSH/p1733105260465839

### DIFF
--- a/ansible/roles/species-list/templates/specieslist-webapp-config.properties
+++ b/ansible/roles/species-list/templates/specieslist-webapp-config.properties
@@ -134,8 +134,8 @@ security.apikey.header.alternatives=Authorization
 #websevice jwt
 webservice.jwt={{ webservice_jwt | default('false') }}
 webservice.jwt-scopes={{ lists_webservice_jwt_scopes | default(webservice_jwt_scopes) | default('') }}
-webservice.client-id={{ (lists_client_id | default(service_client_id)) | default('') }}
-webservice.client-secret={{ (lists_client_secret | default(service_secret)) | default('') }}
+webservice.client-id={{ service_client_id | default('') }}
+webservice.client-secret={{ service_secret | default('') }}
 
 # For backward compatibility:
 userDetails.url={{ specieslist_userdetails_url | default(userdetails_url) | default('https://auth.ala.org.au/userdetails/') }}


joe-lipson (approved):

vjrj:
I introduced (*) this kind of variables (the same in other roles) thinking that will not affect ALA. I don't have access to the slack conversation, but, which is the error? If I'm not wrong, this should use `service_client_id` if `list_client_id` is not present. Sorry for the noise. 

(*) As in the community we use a single inventory for all our services we have to configure different values for each service (in this case for different client ids and secrets that are autogenerated and auto configured in CAS):

image
also other variables different than client-id/secrets:
https://github.com/living-atlases/generator-living-atlas/issues/10

Related:
https://github.com/AtlasOfLivingAustralia/ala-install/pull/855

nickdos:
Hi @vjrj - I should've checked git blame, so my bad for not checking with you about this. 

The issue was that both vars were populated by the inventory and so the code:

```
{{ (lists_client_id | default(service_client_id)) | default('') }}
```

was getting the `lists_client_id` - but we required the `service_client_id` here. We require 2 distinct sets of ID/keys for this app. One for login and another for M2M auth.

So maybe we can un-popuplate the `lists_client_id` value in our inventory file, instead?

Here is our inventory code:

```
    webservice_jwt: true
    webservice_jwt_scopes: ala/internal users/read ala/attrs
    jwt_client_id: "{{ clientId }}"
    jwt_secret: "{{ secret }}"
    lists_client_id: "{{ clientId }}"
    lists_client_secret: "{{ secret }}"
    service_client_id: XXXXXXXXXXXXXXX
    service_secret: "{{ lookup('amazon.aws.aws_secret', 'specieslist-testing.service-secret', nested=true) }}"
```

I've not touched this ansible code in years, so its all new to me.

vjrj:
See how other roles do the same:
```
$ egrep "client|secret|secert" ansible/roles/collectory/templates/config/collectory-config.properties
security.oidc.clientId={{ (collectory_client_id | default(clientId)) | default('') }}
security.oidc.secret={{ (collectory_client_secret | default(secret)) | default('') }}
security.jwt.clientId={{ (collectory_client_id | default(clientId)) | default('') }}
webservice.client-id={{ collectory_client_id | default(webservice_clientId) | default('') }}
webservice.client-secret={{ collectory_client_secret | default(webservice_secert) | default('') }}
```

Forget about `collectory_client_id` and `collectory_client_secret` that I added for the community. But note that we use only one client/secret per application (not two as ALA). 

Then in the ALA collectory inventories you should have:
```
clientId=XXXXXXXXXXXXX
secret=YYYYYYYYYYYYY
webservice_clientId=ZZZZZZZZZZZZZZZZ
webservice_secert=AAAAAAAAAAAAAA
```

What failed with species list? That I introduced the `lists_client_id` (as I did with `collectory_client_id`) without notice that you were using that variable too in:
```
security.jwt.clientId={{ lists_client_id}}
```

So at the end, the same `lists_client_id` is used for all (like in our community). Sorry! 

How to solve this for all without having to change all our inventories in the community (*) ? 
I think we can just restore the previous code like this and modify `security.jwt.clientId` like:
```
security.oidc.clientId={{ (lists_client_id | default(clientId)) | default('') }}
security.oidc.secret={{ (lists_client_secret | default(secret)) | default('') }}
security.jwt.clientId={{ lists_client_id | default(clientId) }} 
webservice.client-id={{ (lists_client_id | default(service_client_id)) | default('') }}
webservice.client-secret={{ (lists_client_secret | default(service_secret)) | default('') }}
```

and remove the `lists_client_id` from the ALA inventories. In other words
- variables used by us: `lists_client_id` and `lists_client_secret`
- variables used by ALA: `clientId`, `secret`, `service_client_id` and `service_secret`.

(*) The file that we use to store passwords and apikeys is created when someone start to create a LA project with random passwords and api keys and we try to not modify this file later. The rest of the inventories are more configurable and we change them on each release I do of the generator following the changes in ala-install and code or the setup introduced in the la-tookit. So we update all our inventories programmatically except the passwords/apikey file.

PR #860, opened 2024-12-04
specieslist-webapp - fix for LA deployments

nickdos:
Revert last change but fix missing var error for `lists_client_id`


vjrj (approved):
LGTM, thanks @nickdos

qifeng-bai (approved):

PR #868, opened 2025-02-26
cas: add option for static .well-known

matthewandrews:
This adds the option for the /cas/oidc/.well-known file to be delivered as a static file by nginx, so that when the CAS app is down, .well-known is still available: this endpoint is required by apps using OIDC, and we've found that .well-known being unavailable can cause apps to fail to start.


ansible/roles/cas5/defaults/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+# By default, CAS serves the response to cas/oidc/.well-known requests from the CAS app itself.
+# If cas_oidc_well_known_static is set to true in an inventory, nginx will serve the response directly from a static file.
+cas_oidc_well_known_static: false
+# OICD scopes for CAS
+cas_oidc_well_known_scopes:
```
sbearcsiro, line 6: 1. If we're defining a var for this, then the list at `cas.authn.oidc.discovery.scopes` in `application.yml` should also use it.
2. This list is missing digivol/internal

ansible/roles/cas5/templates/well-known
```diff
@@ -0,0 +1,186 @@
+{
+  "issuer": "{{auth_cas_url}}/oidc",
+  "scopes_supported": [
+{% for ascope in cas_oidc_well_known_scopes %}
```
sbearcsiro, line 4: We should have an additional loop after this one for the cas_extra_scopes var

ansible/roles/cas5/templates/well-known
```diff
@@ -0,0 +1,186 @@
...
+    "id_token token"
+  ],
+  "subject_types_supported": [
+    "public",
+    "pairwise"
+  ],
+  "claim_types_supported": [
+    "normal"
+  ],
+  "claims_supported": [
```
sbearcsiro, line 21: This is fine but is technically derived from application config (the static app config that's built into the JAR).  Something to keep in mind if the app config is updated.

sbearcsiro (changes requested):
Looks good, just needs consistency around scopes.

ansible/roles/cas5/templates/well-known
matthewandrews: Done in f0a52d99a8f0fe376a710629823fb42a36457d26

ansible/roles/cas5/defaults/main.yml
matthewandrews: 1. done in 57681045977915a718df604f753afb70f5a1cb8a
2. It seems like digivol/internal should be not in the default list - will add in ALA inventory

PR #873, opened 2025-03-17
Feature/tidyup

chrisala:


temi (approved):
just a comment on does internal ip need to set on reporting server to primary server?

chrisala:
> just a comment on does internal ip need to set on reporting server to primary server?

Good catch @temi - it probably should be yes.

PR #874, opened 2025-03-19
Feature/logger jar

yasima-csiro:


ansible/README.md
matthewandrews, line 1: I think this line in the README should change from a tomcat_deploy to an exec_jar

matthewandrews (approved):

ansible/roles/logger-service/templates/logback.xml
```diff
@@ -6,7 +6,7 @@
     <conversionRule conversionWord="wex" converterClass="org.springframework.boot.logging.logback.WhitespaceThrowableProxyConverter" />
 
     <variable name="appName" value="logger-service"/>
-    <variable name="LOG_DIR" value="/var/log/{{ tomcat | default('tomcat9')}}" />
+    <variable name="LOG_DIR" value="{{ jar_log_path | default('/var/log/tomcat9')}}" />
```
sbearcsiro, line 9: IMHO this should retain the tomcat version in the log path for the default case.
ie `{{ jar_log_path | default('/var/log/'  + tomcat)}}` or however you'd express that in ansible

sbearcsiro (changes requested):
Looks good but should retain tomcat version in logger path default

ansible/roles/logger-service/templates/logback.xml
yasima-csiro: Fixed

sbearcsiro (approved):

Issue #854, opened 2024-10-16
Error use_openjdk17 is undefined


vjrj:
After #848 we still get:

```
TASK [java : Show java version] ************************************************
fatal: [ala-install-test-3]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'use_openjdk17' is undefined. 'use_openjdk17' is undefined\n\nThe error appears to be in '/data-additional/jenkins/ala-install-test/ala-install/ansible/roles/java/tasks/main.yml': line 4, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"Show java version\"\n  ^ here\n"}

PLAY RECAP *********************************************************************
ala-install-test-3         : ok=14   changed=0    unreachable=0    failed=1    skipped=16   rescued=0    ignored=0   
```

cc @yasima-csiro

vjrj:
This still fails for some users.

PR #875, opened 2025-03-26
Add option for OpenJDK 17 in defaults

vjrj:
This pull request includes a small change to the `ansible/roles/common/defaults/main.yml` file. The change adds a new configuration option for `use_openjdk17` to avoid errors like #854 and with a similar use than for older versions.

### DIFF
--- a/ansible/roles/common/defaults/main.yml
+++ b/ansible/roles/common/defaults/main.yml
@@ -2,3 +2,4 @@ packaging: "war"
 use_zulu8: False
 use_openjdk8: True
 use_openjdk11: False
+use_openjdk17: False


matthewandrews (approved):

vjrj:
Hi @matthewandrews, I've just added you as reviewer of this. It's quite simple, but newcomers are finding that var as `undefinided` and it's easier to provide a default value -as the rest of java versions-, that to update n inventories.

matthewandrews:
I have merged that now, thanks @vjrj

vjrj:
Thanks indeed!

PR #877, opened 2025-04-03
New bie vars for the LA community

vjrj:
This pull request includes updates to the configuration templates for BIE Hub and BIE Index services. The changes primarily focus on adding more flexibility to the URLs and query contexts by introducing additional default values useful for the LA Community that use unique inventories for all their services.


vjrj:
I'll merge this in short if nobody has a comment. It's add new vars and should not affect to ALA.

PR #881, opened 2025-04-14
Deploy Collectory as systemd service

qifeng-bai:


adam-collins (changes requested):
Remove the duplication by merging the optional exec jar related changes into the `collectory` role.

qifeng-bai:
> Remove the duplication by merging the optional exec jar related changes into the `collectory` role.

I considered that option, but the ‘collectory’ role is also used by our production environment. If we proceed, the next deployment in production might interfere with the existing Tomcat setup.

adam-collins:
Please proceed with the removal of duplication.

qifeng-bai:
> Please proceed with the removal of duplication.

Done

adam-collins (changes requested):
I see that the update made progress by removing the duplication.

It also needs to make the "deploy as service" an option instead of replacing tomcat. Consider this as remaining backwards compatible. The default remains as tomcat.

For a working example, see the `biocache-hub` role. By specifying in the inventory a `packaging` (which triggers the `common` role to set `exec_jar`) and a `classifier`, the biocache-hub role creates a service instead of using tomcat.

ansible/roles/collectory/templates/config/collectory-config.properties
```diff
@@ -143,11 +142,14 @@ gbifApiPassword={{ gbif_api_password | default('') }}
 gbifExportUrlBase={{ gbif_export_url_base | default('') }}
 useGbifDoi={{ gbif_use_doi | default('false') }}
 gbifRegistrationDryRun={{ gbif_registration_dry_run | default('true') }}
+
 {% if collectory_version is version('3.0.0', '<') %}
 gbifLicenceMappingUrl={{ gbif_licence_mapping_url | default('file:///data/ala-collectory/config/default-gbif-licence-mapping.json') }}
 {% else %}
 gbifLicenceMappingUrl={{ gbif_licence_mapping_url | default('file:///data/collectory/config/default-gbif-licence-mapping.json') }}
 {% endif %}
+gbifLicenceMappingUrl={{ gbif_licence_mapping_url | default('file:///data/collectory/config/default-gbif-licence-mapping.json') }}
```
adam-collins, line 151: `gbifLicenceMappingUrl` is already set by the previous `if` block.

ansible/roles/collectory/handlers/main.yml
```diff
@@ -1,10 +1,19 @@
 - name: restart mysql
   service: name={{mysql_service}} state=restarted enabled="yes"
 
+- name: restart collectory
+  service:
+    name: "{{ 'collectory' if exec_jar else tomcat }}"
```
adam-collins, line 6: should this be `collectory_service_system_user`?

ansible/roles/tomcat/tasks/main.yml
```diff
@@ -98,15 +98,15 @@
...
-  - restart tomcat
+#- name: enable the HTTP 1.1 connector (Tomcat-9)
+#  tomcat9_connector: name="HTTP/1.1" bind_addr="{{ tomcat_bind_addr }}" port="{{ tomcat_http_port }}" max_post_size="{{ tomcat_max_post_size }}" relaxed_query_chars="{{ tomcat_relaxed_query_chars }}" tomcat_conf_dir="{{ tomcat_conf_dir }}"
+#  when:
+#    - tomcat_http | bool == True
+#    - tomcat == "tomcat9"
+#  tags:
+#  - tomcat
+#  notify:
+#  - restart tomcat
```
adam-collins, line 109: Revert this change.

ansible/roles/collectory/vars/main.yml
```diff
@@ -1,17 +1,21 @@
...
 artifactId: "{{ collectory_app | default('generic-collectory') }}"
 
 groupId: "{{ collectory_group_id | default('au.org.ala') }}"
 # collectory does not need a value for classifier
 classifier: ''
 
-packaging: "war"
+packaging: "{{ collectory_packaging | default('war') }}"
+
+version: "{{ collectory_version | default('1.5.1') }}"
```
adam-collins, line 16: Revert to the previous default version `LATEST`.

ansible/roles/collectory/tasks/vm-tasks.yml
```diff
@@ -18,21 +18,69 @@
...
+    use_openjdk17: "{{ collectory_version | string is version('5.2.0', '>') }}"
+    log_config_filename: "{{ logback_config_filename }}"
+    java_headless: True
+  tags:
+    - deploy
+    - service
+    - collectory
+  when: exec_jar | default(false) | bool
+
+#todo: check if this is needed
```
adam-collins, line 37: Answering the question, this check (directory exists) is not required because it is performed by the nginx_vhost role.

ansible/roles/collectory/tasks/main.yml
```diff
@@ -44,10 +67,18 @@
     - properties
     - collectory
 
-- name: set data ownership [all data is owned by tomcat]
-  file: path={{ collectory_data_dir }} owner={{ tomcat_user }}  group={{ tomcat_user }}  recurse=true
-  notify: 
-    - restart tomcat  
+#- include_tasks: ../../tomcat_deploy/tasks/main.yml
```
adam-collins, line 70: Remove this unused, commented out, `tomcat_deploy` section.

ansible/roles/collectory/tasks/vm-tasks.yml
```diff
@@ -18,21 +18,69 @@
...
+- name: set log file for collectory systemd service
+  file: path="/var/log/collectory" owner="{{ data_owner_user | default('collectory') }}"  state="directory" group="{{ data_owner_user | default('collectory') }}" recurse=true
+  notify:
+    - restart collectory
+  tags:
+    - collectory
+  when: exec_jar | default(false) | bool
+
+
+#deploy on tomcat
```
adam-collins, line 57: Remove this `set log file for tomcat`. This is done by the tomcat package.

ansible/roles/collectory/tasks/main.yml
```diff
@@ -7,8 +7,30 @@
...
+    msg: >
+      {% if exec_jar | default(false) | bool %}
+        Running as a systemd service.
+      {% else %}
+        Running on {{tomcat}}.
+      {% endif %}
+      Version: {{ version }}.{{ packaging }}
+      Service user: {{ data_owner_user }}
+
+- name: "ensure system user for collectory exists - "
```
adam-collins, line 26: Add a conditional to run this `ensure system user for collectory exists` so it is not run for a tomcat deploy.

adam-collins (changes requested):

ansible/roles/collectory/templates/config/collectory-config.properties
qifeng-bai: removed

ansible/roles/collectory/handlers/main.yml
qifeng-bai: Done

ansible/roles/tomcat/tasks/main.yml
qifeng-bai: Done

ansible/roles/collectory/tasks/vm-tasks.yml
qifeng-bai: removed

ansible/roles/collectory/tasks/main.yml
qifeng-bai: removed

ansible/roles/collectory/tasks/main.yml
qifeng-bai: It sets on line 11: data_owner_user: "{{ collectory_service_system_user if exec_jar else tomcat_user }}"

adam-collins: The reason for the conditional is because I do not trust ansible to create the tomcat user in an identical fashion to the tomcat package installer. While the tomcat user creation should take place before this task it is not certain.

qifeng-bai: Done

ansible/roles/collectory/tasks/vm-tasks.yml
```diff
@@ -18,21 +18,75 @@
...
+    use_openjdk17: "{{ collectory_version | string is version('5.2.0', '>') }}"
+    log_config_filename: "{{ logback_config_filename }}"
+    java_headless: True
+  tags:
+    - deploy
+    - service
+    - collectory
+  when: exec_jar | default(false) | bool
+
+- name: Extract assets (images) from jar for systemd service
```
adam-collins, line 37: If `Extract assets (images) from jar for systemd service` needs to be removed. This is only required if the collectory jar is not built correct. Fix it in the collectory build process.

ansible/roles/collectory/tasks/vm-tasks.yml
```diff
@@ -18,21 +18,75 @@
...
   tags:
     - properties
     - collectory
+  when: not (exec_jar | default(false) | bool)
 
 - include_tasks: ../../tomcat_deploy/tasks/main.yml
   vars:
-    war_url: '{{ collectory_war_url }}'
+    #only used when tomcat_deploy_from_url
+    war_url: '{{ collectory_war_url | default('')}}'
```
adam-collins, line 84: The `collectory_war_url` was previously set in `vars/main.yml` as 
`collectory_war_url: "{{ collectory_maven_repo_ws_url | default(maven_repo_ws_url) }}"`

Please revert to the prior default value instead of `empty string`.

ansible/roles/tomcat/tasks/main.yml
```diff
@@ -98,15 +98,15 @@
   notify:
   - restart tomcat
 
-- name: enable the HTTP 1.1 connector (Tomcat-9)
```
adam-collins, line 101: This section `enable the HTTP 1.1 connector (Tomcat-9)` is still disabled. Please revert this change. Make it conditional to exec_jar if requried.

adam-collins (changes requested):
Changes requested.

ansible/roles/tomcat/tasks/main.yml
adam-collins: I was unaware of any new issue with this role as I have recently used it on new Ubuntu 22 servers without error. While still avoiding fixing the problem, a less severe workaround is to use `ignore_errors: yes`.

qifeng-bai: The task fails with the error /usr/bin/env: ‘python3’: No such file or directory, despite python3 being installed and accessible. We can successfully run python3 from the terminal when logged in as the tomcat user.

ansible/roles/collectory/vars/main.yml
qifeng-bai: done

ansible/roles/tomcat/tasks/main.yml
adam-collins: If I understand correctly, this is running on an older server, so I expect it might well be only specific to that one server since it does run without error in other environments. It is very odd that the other tomcat9_connector tasks run without this same error. 

My intention is to not touch something that is working, and this task is definitely working in some environments. 

While it is not ideal, `ignore_errors: yes` can be used if the environment where it is not working has that connector already configured in tomcat.

qifeng-bai: I have added 'ignore-error:yes'

ansible/roles/collectory/vars/main.yml
```diff
@@ -1,17 +1,21 @@
...
 groupId: "{{ collectory_group_id | default('au.org.ala') }}"
 # collectory does not need a value for classifier
 classifier: ''
 
-packaging: "war"
+packaging: "{{ collectory_packaging | default('war') }}"
+
+version: "{{ collectory_version | default('LATEST') }}"
 
-collectory_war_url: "{{ collectory_maven_repo_ws_url | default(maven_repo_ws_url) }}"
```
adam-collins, line 14: Revert this line so that `collectory_war_url` is always set to the original value.

adam-collins (changes requested):
One more change requested.

ansible/roles/collectory/vars/main.yml
qifeng-bai: done

qifeng-bai: done

adam-collins (approved):
approved

Issue #771, opened 2024-01-23
sensitive-data-service.yml issues


adam-collins:
Failed to run sensitive-data-service.yml on a new server.

Issues found
- cannot override namematching-service version. The default version (latest) fails. Success with ala_namematching_service_version=1.8-SNAPSHOT+0~20220601053108.49~1.gbp224ac7
- need directory /var/log/atlas
- need directory /var/log/atlas/sensitive-data-service/ as user sds-data
- edit /data/ala-sensitive-data-service/config/config.yml for the installed Lucene directory /data/lucene/namematching-sd
- edit /etc/systemd/system/multi-user.target.wants/ala-namematching-service.service for java arm64 instead of amd64
- edit /etc/systemd/system/multi-user.target.wants/ala-sensitive-data-service.service for java arm64 instead of amd64
- ran it with deployment_type=vm in the inventory

adam-collins:
Can anyone explain to me why sensitive-data-service requires namematching-service installed? I cannot find how https://github.com/AtlasOfLivingAustralia/ala-sensitive-data-service uses https://github.com/AtlasOfLivingAustralia/ala-namematching-service.

vjrj:
> Can anyone explain to me why sensitive-data-service requires namematching-service installed? 

Sorry, this is my fault. I think that because sensitive-data-service needs a working name index, see step two of:
https://github.com/AtlasOfLivingAustralia/ala-sensitive-data-service/blob/master/README.md
and to avoid duplicate the name index download code that was in both deb packages, I opted to install directly the name-matching service as dependency.

It would have been better to have a package or task just to download the name index, shared by both services and that the services role only installed their executable, but already with the necessary data in place.

nickdos:
> I've commented out the name-matching references and running on test server, install ran to completion without errors.

@adam-collins pointed out that the Lucene index is now not installed but is needed. Reason it didn't fail in test, was likely due to the index still being present from before the change (but not removed). I'll attempt to extract the index step, as suggested by @vjrj .

nickdos:
[PR updated](https://github.com/AtlasOfLivingAustralia/ala-install/pull/890).

PR #890, opened 2025-06-05
#771 Removed refs to name-matching-service from sensitive-data-servic…

nickdos:
#771


nickdos:
Debated whether to remove or comment-out. Opted for the latter, in case we need to back it out.

adam-collins (approved):
approved

Issue #895, opened 2025-06-17
Make all vars compatible with ansible `2.8`.


vjrj:
Follow work of #892 for make compatible vars for ansible `2.8`.

Currently, in:
```
$ pcregrep -r -M "^\s*vars:\s*\n\s*- " ansible   | cut -d ":" -f 1
(...)
ansible/solrcloud-docker.yml
    - deployment_type
ansible/biocache-hub-standalone-docker.yml
    - deployment_type
ansible/roles/userdetails/tasks/docker-tasks.yml
    - cas_db_hostname
ansible/roles/userdetails/tasks/docker-tasks.yml
    - cas_db_hostname
ansible/species-list-docker.yml
    - deployment_type
ansible/biocache-service-clusterdb-docker.yml
    - deployment_type
ansible/bie-hub-docker.yml
    - deployment_type
ansible/namematching-service-docker.yml
    - deployment_type
ansible/portainer.yml
    - deployment_type
ansible/sensitive-data-service-docker.yml
    - deployment_type
ansible/auth-docker.yml
    - deployment_type
ansible/gatus.yml
    - deployment_type
ansible/bie-index-docker.yml
    - deployment_type
ansible/cassandra-docker.yml
    - deployment_type
```

PR #896, opened 2025-06-17
All vars compatible with ansible 2.8+

vjrj:
Fix for #895 continuing the work of #892.

PS: It affects to roles not used by ALA so it should be quite secure to merge it.


matthewandrews (approved):
LGTM

PR #905, opened 2025-07-15
support [ and ] in URLs for biocache-service as exec-jar

adam-collins:


adam-collins:
wrong commit text, should be biocache-hub

PR #913, opened 2025-08-10
Update nginx PPA key in setup-ubuntu.yml

ezaton:
The nginx keyserver id has changed to 2FD21310B49F6B46 - or so it claims in the logs.

### DIFF
--- a/ansible/roles/nginx/tasks/setup-ubuntu.yml
+++ b/ansible/roles/nginx/tasks/setup-ubuntu.yml
@@ -11,7 +11,7 @@
 - name: Add PPA key for Nginx repository
   apt_key:
     keyserver: "hkp://keyserver.ubuntu.com:80"
-    id: "ABF5BD827BD9BF62"
+    id: "2FD21310B49F6B46"
   when:
     - nginx_ppa_use is defined and nginx_ppa_use | bool == True
 


vjrj (changes requested):
Sorry, I don't find this new key. 

And also the previous key is not expired in my side:

```
~# apt-key  list
/etc/apt/trusted.gpg
--------------------
pub   rsa2048 2011-08-19 [SC] [expires: 2027-05-24]
      573B FD6B 3D8F BC64 1079  A6AB ABF5 BD82 7BD9 BF62
uid           [ unknown] nginx signing key <signing-key@nginx.com>
```

If you want to improve this part, I think that you should follow this:
- https://blog.nginx.org/blog/updating-pgp-key-for-nginx-software
and try to not use `apt_key` as is deprecated in debian/ubuntu :    
  - https://github.com/ansible/ansible/issues/78063
 - https://www.jeffgeerling.com/blog/2022/aptkey-deprecated-debianubuntu-how-fix-ansible

ezaton:
Hi. Running a vanilla deployment using ALA toolkit (the docker container, answering questions and so on), failed the initial run with the following error:
 ```
 Task:  nginx : Add PPA for Nginx on current versions failed with message: 
Failed to update apt cache: W:GPG error: http://nginx.org/packages/mainline/ubuntu jammy InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 2FD21310B49F6B46, E:The repository 'http://nginx.org/packages/mainline/ubuntu jammy InRelease' is not signed.
``` 
I had to login manually to the servers and import the said public key, which is different than what is provided in the Ansible task.
The system is Ubuntu 22.04 on AWS in Ireland zone. Public IP addresses - vanilla deployment of Ubuntu server from EC2 templates.

vjrj:
Thanks for the report, I did a new PR with the suggestion I did you in my review: https://github.com/AtlasOfLivingAustralia/ala-install/pull/915

Issue #975, opened 2026-07-09
libvips42t64 not available in u22


vjrj:
I get this error installing in u22:

> TASK [image-service : Install libvips] *****************************************
fatal: [ala-install-test-2.images]: FAILED! => {"changed": false, "msg": "No package matching 'libvips42t64' is available"}

But `libvips42` is present.

PR #976, opened 2026-07-10
libvips_packages for ubuntu < 24.04

vjrj:
Fix for #975.


ansible/roles/image-service/vars/Debian.yml
```diff
@@ -10,11 +10,15 @@ image_optimisation_packages:
   - gifsicle
   - webp
   - libavif-bin
+# On Ubuntu < 24.04 (e.g. 22.04) these libraries predate the 64-bit time_t (t64)
+# rename and the libtiff6 SONAME bump, so use the legacy package names there.
+# Ubuntu 24.04+ (and other distros) keep the current names. See issue #975.
+_ubuntu_pre_t64: "{{ ansible_distribution == 'Ubuntu' and ansible_distribution_version is version('24.04', '<') }}"
```
sbearcsiro: I'm not a fan of adding logic inside the vars files *and* this feels a little to specific... so either can we add something like:

  with_first_found:
    - "vars/{{ ansible_distribution }}-{{ ansible_distribution_version }}.yaml"
    - "vars/{{ ansible_distribution }}.yaml"
to the tasks yml

and then add a specific 22.04.yaml file?

OR

discover t64 support at runtime?

- name: Check if t64 packages are standard on this system
  command: apt-cache show libvips42t64
  register: t64_check
  failed_when: false
  changed_when: false

... there are obvious problems with both approaches so feel free to add your own...

sbearcsiro (changes requested):

ansible/roles/image-service/vars/Debian.yml
vjrj: I implemented and tested the snd option.

Issue #977, opened 2026-07-13
image-service: optional svgo (snap) install aborts the whole deploy where snapd is unavailable


vjrj:
The `image-service` role installs `svgo` via `snap` on Ubuntu. On hosts where `snapd` cannot mount squashfs images (in our case unprivileged LXC containers in Proxmox, etc.), but also minimal/containerised hosts, this task fails, even though SVG optimisation is an optional enhancement.

PR #978, opened 2026-07-13
fix(image-service): install svgo via npm where snap can't mount squashfs

vjrj:
svgo installs via snap on Ubuntu, but snapd can't mount its squashfs image inside unprivileged containers (LXC/Docker), so the task fails with "system does not fully support snapd: cannot mount squashfs" and aborts the service deploy, even though SVG optimisation is optional.

Detect that case via ansible_virtualization_type (lxc/docker/podman/container/ systemd-nspawn/openvz) and install svgo with npm there instead (ensuring npm is present first, since the image is not otherwise installed). 

Tested on Ubuntu 22.04 in an unprivileged Proxmox LXC (ala-install-deploy-tests #1409): snap task skips, npm+svgo install cleanly, deploy completes green.

Fix for #977.

### DIFF
--- a/ansible/roles/image-service/tasks/main.yml
+++ b/ansible/roles/image-service/tasks/main.yml
@@ -196,20 +196,34 @@
         state: present
       become: true
 
+    # Snap has to mount a squashfs image, which is not permitted inside
+    # unprivileged containers (LXC, Docker, ...): snap install aborts with
+    # "system does not fully support snapd: cannot mount squashfs". Detect that
+    # case up front and install svgo (a plain Node CLI) via npm instead.
     - name: Install svgo for SVG optimisation (Snap)
       snap:
         name: svgo
         state: present
       become: true
-      when: ansible_distribution == "Ubuntu"
+      when:
+        - ansible_distribution == "Ubuntu"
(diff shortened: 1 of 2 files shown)


ansible/roles/image-service/tasks/main.yml
```diff
@@ -196,20 +196,34 @@
...
+    # case up front and install svgo (a plain Node CLI) via npm instead.
     - name: Install svgo for SVG optimisation (Snap)
       snap:
         name: svgo
         state: present
       become: true
-      when: ansible_distribution == "Ubuntu"
+      when:
+        - ansible_distribution == "Ubuntu"
+        - ansible_virtualization_type not in ['lxc', 'docker', 'podman', 'container', 'systemd-nspawn', 'openvz']
```
sbearcsiro: can we register this in a var so that it's not repeated?

sbearcsiro (approved):
Looks good but see comment

ansible/roles/image-service/tasks/main.yml
vjrj: I did it, and test it.

PR #989, opened 2026-08-19
Derive ALLOWED_SENDER_DOMAINS from the deployment's own sender

vjrj:
`ALLOWED_SENDER_DOMAINS` for the postfix container defaulted to the literal `l-a.site`, and `email_allowed_domains` is not set anywhere in this repo, so the default always applied: every deployment authorised a domain belonging to someone else. Postfix reports it at startup as `Setting up allowed SENDER domains: l-a.site`.

It is now derived from `email_sender`, which the same block already uses for `RELAYHOST_USERNAME`. When `email_sender` is unset the result is empty, which authorises nobody. That is a safer failure than authorising a third party's domain.


vjrj:
I merge this as this is no in use by ALA and it's well tested in our side.

PR #993, opened 2026-08-20
Gatus: configurable up conditions, a stricter root check, and skip unprobeable paths

vjrj:
The `gatus` role generates one check per `nginx_paths` entry of each vhost. That list describes how nginx routes requests, not which URLs are health endpoints, and the mismatch produces checks that cannot report anything useful. Three changes.

**Entries that are not URLs at all.** `nginx_paths` carries nginx location modifiers (`~`, `~*`, `=`, `^~`, `@`): `sensitive-data-service` declares `~* /location/`, which rendered into a check named `https://<host>/~* -location` that can only ever time out. Entries with `disable: true` are not served either, since `fragment_73_location.j2` skips them. Both are skipped now.

**A 404 on the vhost root reported as healthy.** `[STATUS] == 200` is too strict for the rest of `nginx_paths`, since many API paths answer 400/404 without parameters while perfectly healthy, so `gatus_up_condition` / `gatus_up_conditions` let an inventory relax it. The root cannot be relaxed the same way: a 404 on `/` means no `server_name` matched and the request reached the wrong machine, and under a relaxed condition every check in that group then reports green. The root gets its own `gatus_root_up_condition`, defaulting to `[STATUS] < 400`.

Whatever is configured is always paired with `[CONNECTED] == true`, because a status-only condition is *satisfied* when Gatus never reached the host at all: a failed request leaves the status at 0. `gatus_up_conditions` (plural) replaces the whole pair for callers that want full control, and wins for the root too.

**Container support.** The data directory task is split so container deployments use the `data_dir` layout. The existing task is gated `deployment_type != 'container'`, so every other deployment is untouched.


vjrj:
Merged as this does not affect ALA.  

For the record our CI builds this: https://gatus.l-a.site/ with this rol if someone is interested.

PR #996, opened 2026-08-21
Make the pipelines filesystem backend configurable

vjrj:
`la-pipelines-local.yaml` hardcodes `hdfs://{{ hadoop_master }}:{{ hadoop_port }}` in front of every path it renders. That is right for a VM or cluster install, and wrong for a single-node one where the pipelines run against the local filesystem: there is no HDFS to point at.

The prefix now comes from one variable. `pipelines_use_hdfs` defaults to `deployment_type | default('vm') != 'container'`, so a VM install renders exactly what it renders today, including an inventory that never sets `deployment_type` at all. I checked rather than assumed, by rendering the template both ways and diffing: byte-identical with the variable unset and with it set to `vm`; in container mode the `hdfs://` prefixes drop out and nothing else moves.

The host-provisioning tasks in `tasks/main.yml` (docker, yq, docopts, the HDFS directories, the systemd units) are gated the same way rather than removed, so a VM install still runs all of them.

### DIFF
--- a/ansible/roles/pipelines/tasks/main.yml
+++ b/ansible/roles/pipelines/tasks/main.yml
@@ -7,8 +7,11 @@
     - properties
     - la-pipelines-config
 
+# Host-provisioning tasks below are skipped in container deployments (the pipelines
+# image ships these dependencies); on VMs they run exactly as before.
 - name: Install aptitude using apt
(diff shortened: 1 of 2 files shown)


matthewandrews (approved):

vjrj:
thanks @matthewandrews

PR #998, opened 2026-08-22
Write the gatus check on the host that runs gatus

vjrj:
`nginx_vhost` renders a gatus check per application into `gatus_config_dir`, so an application appears on the dashboard without anyone maintaining the gatus config by hand for this kind of check.

It writes that file on the host being configured. gatus reads only the config directory of the machine it runs on, and hosts do not share a filesystem, so unless gatus sits on the application's host the file lands where nothing reads it. Nothing fails, the check just never exists.

Delegate the copy to the host in the `gatus` group and resolve `gatus_config_dir` from that host's vars. Also guard on the group being non-empty, since `use_gatus` with no gatus host aborts the play on `groups['gatus'][0]`, and drop the `with_items` loop, since `http.j2` never references `item` and the loop rendered the same file once per path.

Single host installs are unaffected: the file ends up where it does today.

### DIFF
--- a/ansible/roles/nginx_vhost/tasks/main.yml
+++ b/ansible/roles/nginx_vhost/tasks/main.yml
@@ -153,13 +153,18 @@
   pause: prompt="nginx_paths value is {{nginx_paths}}"
   when: nginx_paths_debug_enabled is defined and nginx_paths_debug_enabled | bool == True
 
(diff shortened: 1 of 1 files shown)


vjrj:
Merged as gatus is not used by ALA

PR #1008, opened 2026-09-08
Optional `carto_api_key` for the bie-hub and biocache-hub basemaps

vjrj:
CARTO basemaps now require an API key, so the tiles both hubs ship with come back stamped "API KEY REQUIRED". Reported by Fernando Belmonte (SiBBr); Nick dos Remedios confirmed ALA hit the same thing and fixed it by overriding the full tile URL with the key appended. Both said a dedicated variable would be better. Slack thread:

https://atlaslivingaustralia.slack.com/archives/CCTFGEU1G/p1788382741261379

Today the only way to supply a key is to retype the whole tile URL with the key glued on, once per service, in every inventory. Now:

```ini
[all:vars]
carto_api_key = your-carto-key
```

Also adds `carto_basemap_url` (change the style, still get the key appended) and accepts `map_minimal_url` as an alias for the historical `map_mininal_url` typo.

The default tile host moves from the deprecated `cartodb-basemaps-{s}.global.ssl.fastly.net` to the current `basemaps.cartocdn.com`. Both return the identical watermarked PNG when no key is present (18422 bytes, checked with `curl`), so only the key changes anything visible.

## Existing inventories are untouched

`default_map_url` and `map_mininal_url` still override everything, query string included. `carto_api_key` is only appended to the value the role itself computes, never to one that came from an inventory, so an inventory already carrying `?key=...` cannot end up with two. ALA and SiBBr need no changes on their side.

Verified per case with real `defaults/` precedence: nothing set, key only, full URL with `?key=` (the ALA config), that same URL plus `carto_api_key` set, only one of the two URLs overridden, the correctly spelled alias, and an empty key. Also rendered end to end through `la-docker-compose`, both the deploy path and its offline CI drift check (`config baseline OK: 11 files, 374 URL keys unchanged`).

The composition lives in the templates rather than in `defaults/main.yml` on purpose: offline config renderers load the defaults standalone and merge them with `combine()`, where a default cannot reference a sibling default.


ansible/roles/biocache-hub/templates/config/config.properties
```diff
@@ -152,7 +152,7 @@ map.zoomOutsideScopedRegion={{zoom_outside_scoped_region|default('true')}}
 map.height={{map_height|default('600')}}
 geocode.region={{geocode_region | default('AU')}}
 
-map.minimal.url={{ map_mininal_url | default('https://cartodb-basemaps-{s}.global.ssl.fastly.net/light_all/{z}/{x}/{y}.png') }}
+map.minimal.url={{ map_mininal_url | default(map_minimal_url | default(carto_basemap_url ~ (('?key=' ~ carto_api_key) if carto_api_key else ''))) }}
```
nickdos, line 155: I think this is probably an existing bug but it looks like a typo for `map_mininal_url` - should it be `map_minimal_url`?
AI suggests it should be:
`map.minimal.url={{ map_minimal_url | default(carto_basemap_url ~ (carto_api_key | ternary('?key=' ~ carto_api_key, ''))) }}`
I've not tested it though.

nickdos: Did a [search](https://github.com/search?q=org%3AAtlasOfLivingAustralia+map_mininal_url&type=code) for `map_mininal_url` and its found in lots of files, so leave it as it was. Seems like a typo but not 100% sure now.

nickdos (approved):

